PR #11, opened 2023-01-26
Set up CMake build

yim-lee:
Depends on https://github.com/apple/swift-asn1/pull/9 and https://github.com/apple/swift-crypto/pull/144


yim-lee:
Ah, I deleted my fork forgetting that this hasn't been merged yet. Will open another.

PR #23, opened 2023-02-22
CMS Signing support

Lukasa:
This patch adds support for basic CMS operations. We aren't sure that the API surface here is right, so for now we're going to provide this under an SPI guard so that we can make any amendments we might need.

### DIFF
--- a/Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
+++ b/Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
@@ -20,6 +20,7 @@ public enum CMS {
     public static func sign<Bytes: DataProtocol>(
         _ bytes: Bytes,
         signatureAlgorithm: Certificate.SignatureAlgorithm,
+        additionalCertificates: [Certificate] = [],
         certificate: Certificate,
         privateKey: Certificate.PrivateKey
     ) throws -> [UInt8] {
@@ -27,6 +28,7 @@ public enum CMS {
         return try sign(
             signatureBytes: ASN1OctetString(signature),
             signatureAlgorithm: signatureAlgorithm,
+            additionalCertificates: additionalCertificates,
             certificate: certificate
         )
     }
@@ -36,9 +38,15 @@ public enum CMS {
     public static func sign(
         signatureBytes: ASN1OctetString,
         signatureAlgorithm: Certificate.SignatureAlgorithm,
+        additionalCertificates: [Certificate] = [],
         certificate: Certificate
     ) throws -> [UInt8] {
-        let signedData = try self.generateSignedData(signatureBytes: signatureBytes, signatureAlgorithm: signatureAlgorithm, certificate: certificate)
+        let signedData = try self.generateSignedData(
+            signatureBytes: signatureBytes,
+            signatureAlgorithm: signatureAlgorithm,
+            additionalCertificates: additionalCertificates,
+            certificate: certificate
+        )
 
         var serializer = DER.Serializer()
         try serializer.serialize(signedData)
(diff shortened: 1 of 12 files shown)


Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
```diff
@@ -0,0 +1,222 @@
...
+            // This force-unwrap is safe: we know there are certificates because we've located at least one certificate from this set!
+            let untrustedIntermediates = CertificateStore(signedData.certificates!)
+            var verifier = Verifier(rootCertificates: trustRoots, policy: policy)
+            let result = await verifier.validate(leafCertificate: signingCert, intermediates: untrustedIntermediates)
+
+            switch result {
+            case .validCertificate:
+                return .validSignature(.init(signer: signingCert))
+            case .couldNotValidate(let validationFailures):
+                return .unableToValidateSigner(.init(validationFailures: validationFailures))
```
yim-lee, line 146: Can we also include `signingCert` in `unableToValidateSigner`?

Lukasa: Done.

Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
```diff
@@ -0,0 +1,240 @@
...
+    @_spi(CMS)
+    public enum Error: Swift.Error {
+        case incorrectCMSVersionUsed
+    }
+
+    @_spi(CMS)
+    public enum SignatureVerificationResult {
+        case validSignature(Valid)
+        case unableToValidateSigner(SignerValidationFailure)
+        case invalidCMSBlock(InvalidCMSBlock)
```
dnadoba, line 177: This looks like a good use case for a typed error combined with `Result` e.g.:
```suggestion
    public typealias SignatureVerificationResult = Result<Valid, VerificationError>
    public enum VerificationError: Error {
        case unableToValidateSigner(SignerValidationFailure)
        case invalidCMSBlock(InvalidCMSBlock)
```

Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
```diff
@@ -0,0 +1,240 @@
...
+    func certificate(signerInfo: CMSSignerInfo) throws -> Certificate? {
+        switch signerInfo.signerIdentifier {
+        case .issuerAndSerialNumber(let issuerAndSerialNumber):
+            for cert in self {
+                if cert.issuer == issuerAndSerialNumber.issuer && cert.serialNumber == issuerAndSerialNumber.serialNumber {
+                    return cert
+                }
+            }
+        case .subjectKeyIdentifier:
+            // This is unsupported for now.
```
dnadoba, line 227: Should we throw instead for now?

Sources/X509/CryptographicMessageSyntax/CMSSignerInfo.swift
```diff
@@ -33,20 +33,25 @@ import SwiftASN1
...
-    var signatureAlgorithm: AlgorithmIdentifier
-    var signature: ASN1OctetString
-    
+    @usableFromInline var version: CMSVersion
+    @usableFromInline var signerIdentifier: CMSSignerIdentifier
+    @usableFromInline var digestAlgorithm: AlgorithmIdentifier
+    @usableFromInline var signatureAlgorithm: AlgorithmIdentifier
+    @usableFromInline var signature: ASN1OctetString
+
+    @inlinable
```
dnadoba, line 54: Do we really need `@inlinable` here or is `@usableFromInline` enough?  AFAICT nothing is generic and therefore doesn't really benefit from `@inlinable`

Lukasa: Having everything be `@inlinable` is still useful, even without generics: it allows the compiler to observe that these functions don't affect global state, causing ARC traffic to be reduced.

Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
Lukasa: We will, because we'll fail to find a signer.

dnadoba (approved):
Great! LGTM now

Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
yim-lee: 🙏

PR #24, opened 2023-03-01
Handle leaves as trust roots

Lukasa:
In some cases, we will have the leaf certificate in the trust root store. In this case, we can immediately pass that on to policy verification.


dnadoba (approved):
LGTM

PR #29, opened 2023-03-02
Setup CMake build

yim-lee:


yim-lee:
@compnerd We are going to add this new dependency to SwiftPM

tomerd (approved):

PR #31, opened 2023-03-02
Explicitly import _Concurrency

yim-lee:
Motivation:
Swift CI disables implicit concurrency, so build fails without the import.

Modification:
Add explicit import for `_Concurrency`.


tomerd (approved):

tomerd:
@Lukasa we can discuss why this is needed

PR #34, opened 2023-03-06
Add initial RFC5280 policy.

Lukasa:
This is the first of a series of PRs that adds support for RFC 5280 policies as a single verifier policy. For now we take the low hanging fruit, expiry.

### DIFF
--- a/Sources/X509/Verifier/RFC5280/RFC5280Policy.swift
+++ b/Sources/X509/Verifier/RFC5280/RFC5280Policy.swift
@@ -21,37 +21,17 @@ import Foundation
 /// 1. Expiry. Expired certificates are rejected.
 public struct RFC5280Policy: VerifierPolicy {
     @usableFromInline
-    let validationTime: Date
+    let expiryPolicy: ExpiryPolicy
 
     @inlinable
     public init(validationTime: Date) {
-        self.validationTime = validationTime
+        self.expiryPolicy = ExpiryPolicy(validationTime: validationTime)
     }
 
     @inlinable
     public func chainMeetsPolicyRequirements(chain: UnverifiedCertificateChain) -> PolicyEvaluationResult {
-        if case .failsToMeetPolicy(let reason) = self._validateExpiry(chain) {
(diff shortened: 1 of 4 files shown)


Sources/X509/Verifier/RFC5280/RFC5280Policy.swift
```diff
@@ -0,0 +1,59 @@
...
+                return .failsToMeetPolicy(reason: "RFC5280Policy: Certificate \(cert) is not yet valid")
+            }
+
+            if self.validationTime > cert.notValidAfter {
+                return .failsToMeetPolicy(reason: "RFC5280Policy: Certificate \(cert) has expired")
+            }
+        }
+
+        return .meetsPolicy
+    }
```
dnadoba, line 58: I'm wondering if we should split this up into its own `internal` policy and make `RFC5280Policy` be a struct which holds a concrete instance of it. This way we can test the expiry policy in isolation and shouldn't hit any performance degradation because we hold a concrete instance in the public `RFC5280Policy` type (contrary to the non-variadic version of the current `PolicySet` wrapper).

The reason I would like us to do this is because we never actually check the returned failure reason. If we later add more verifiers and just check if verification fails we can't easily verify if the policy we wanted to test has actually triggered the failure or a completely unrelated policy. 

Alternatively we could also add some kind of error code to `.failsToMeetPolicy` which we can then check in tests.

Lukasa: I've been reluctant to add an error code to `.failsToMeetPolicy` because it doesn't compose very well: it's hard to avoid collisions of error codes in a graceful way.

I do quite like the idea of splitting this up into a little mini policy set, so we can do that.

Lukasa: Ok, that's done: I've duplicated all the tests with a common body so that we can confirm both that the actual policy we expect to reject this does, _and_ that the meta policy rejects it as well. That acts as a kind of integration test for us.

dnadoba (approved):
Great! Thanks a lot!

PR #35, opened 2023-03-08
Avoid re-serializing certificates

Lukasa:
When we decode a certificate, we produce an in-memory representation. This representation is something we are capable of encoding. If we re-encode that directly, we can produce an output that is different to the one we injested. This happens when there are objects that can be canonicalised, such as replacing an explicit ASN1 NULL with an absent field where both are considered equivalent.

This patch rewrites the code to ensure that we preserve the bytes of the certificate that we deserialized. This allows us to write them back unchanged, ensuring that certificates can round-trip through this library without modification.

A side-effect of this change is that we need to essentially encode the certificate on construction whenever we build them in memory. That's not a big deal, but it is a bit of an annoyance.

### DIFF
--- a/Sources/X509/Certificate.swift
+++ b/Sources/X509/Certificate.swift
@@ -133,6 +133,20 @@ public struct Certificate {
     /// The signature algorithm used to produce ``signature-swift.property``.
     public let signatureAlgorithm: SignatureAlgorithm
 
+    /// The bytes of the ``Signature``.
+    ///
+    /// These are preserved to ensure that we reserialize exactly what we deserialized, regardless
+    /// of any canonicalisation we might do.
+    @usableFromInline
+    internal let signatureBytes: ArraySlice<UInt8>
+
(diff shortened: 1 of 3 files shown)


Sources/X509/Certificate.swift
```diff
@@ -263,22 +285,41 @@ extension Certificate: DERImplicitlyTaggable {
...
+            coder.serializeRawBytes(self.tbsCertificateBytes)
+            coder.serializeRawBytes(self.signatureAlgorithmBytes)
+            coder.serializeRawBytes(self.signatureBytes)
         }
     }
 }
+
+extension DER.Serializer {
+    @inlinable
+    static func serialized<Element: DERSerializable>(element: Element) throws -> ArraySlice<UInt8> {
```
dnadoba, line 320: That is actually a quite common operation. Might be worth to put this into `swift-asn1` in the future

dnadoba (approved):
LGTM

PR #36, opened 2023-03-08
Add BasicConstraints enforcement to RF5280Policy

Lukasa:
This patch adds support for enforcing BasicConstraints in the RFC 5280 policy.


FranzBusch (approved):
LGTM!

PR #38, opened 2023-03-14
Add support for IP address constraints.

Lukasa:
This patch adds more partial name constraints support, this time for IP address constraints.

### DIFF
--- a/Sources/X509/Verifier/RFC5280/IPConstraints.swift
+++ b/Sources/X509/Verifier/RFC5280/IPConstraints.swift
@@ -1,0 +1,98 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftCertificates open source project
+//
+// Copyright (c) 2023 Apple Inc. and the SwiftCertificates project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftCertificates project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftASN1
+
+extension NameConstraintsPolicy {
+    /// Validates that an IP address matches a constraint.
+    ///
+    /// The rules for IP address constraints are fairly simple. The constraint contains both a subnet
+    /// and a subnet mask, while the `ipAddress` will contain just the bytes of the address. A constraint
+    /// matches if the address is part of the subnet defined by the mask.
+    ///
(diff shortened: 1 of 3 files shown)


FranzBusch (approved):
LGTM!

Sources/X509/Verifier/RFC5280/IPConstraints.swift
```diff
@@ -0,0 +1,98 @@
...
+
+        // All remaining bytes must be zero.
+        let nextIndex = self.index(after: firstInterestingIndex)
+        return self[nextIndex...].allSatisfy { $0 == 0 }
+    }
+
+    @inlinable
+    subscript(offset offset: Int) -> UInt8 {
+        return self[self.startIndex + offset]
+    }
```
dnadoba, line 76: This will be available in the whole target and we can't make it file private. Should we move it to a separate file? (honest question)

Sources/X509/Verifier/RFC5280/IPConstraints.swift
```diff
@@ -0,0 +1,98 @@
...
+            // IPv6
+            return addressIsInSubnet(address: ipAddress.bytes, subnet: constraint.bytes)
+        default:
+            // No match or an invalid format.
+            return false
+        }
+    }
+}
+
+extension ArraySlice where Element == UInt8 {
```
dnadoba, line 42: Nit
```suggestion
extension ArraySlice<UInt8> {
```

Sources/X509/Verifier/RFC5280/IPConstraints.swift
```diff
@@ -0,0 +1,98 @@
...
+    @inlinable
+    var isValidCIDRMask: Bool {
```
dnadoba, line 44: I'm wondering what I general rule of thumb is on when we want to add functionally as an extension like this or declare it to as a free or static function.

I personally like these extensions, it's just a pity that we can't make them `fileprivate`. For that reason I often fall back to static functions to restrict its visibility to a related namespace.

Sources/X509/Verifier/RFC5280/IPConstraints.swift
Lukasa: I'm inclined to leave it. It's not incorrect to use in general.

Sources/X509/Verifier/RFC5280/IPConstraints.swift
Lukasa: Yeah, I'm largely not too worried about it here, but I get the concern. For larger projects I think it would be a bigger deal, but in this instance I think we're probably safe.

dnadoba (approved):

PR #39, opened 2023-03-14
Update Swift Crypto and fix RSA keys

Lukasa:
This updates Swift Crypto to 2.4.0 and fixes RSA key serialization on all platforms. It also adds a defensive check that we only accept appropriately-formatted RSA keys, and some regression tests.


Lukasa:

PR #41, opened 2023-03-14
Improve our handling of trailing periods.

Lukasa:
We incorrectly considered a trailing period to be a match-all clause. It's not. Make the code a little less clever to compensate.

### DIFF
--- a/Tests/X509Tests/DNSNamesTests.swift
+++ b/Tests/X509Tests/DNSNamesTests.swift
@@ -232,6 +232,10 @@ final class DNSNamesTests: XCTestCase {
             ("1234567.example.com", "example.com", true),
             ("foo.1234567.example.com", "foo.1234567.example.com", true),
             ("foo.example.123", "foo.example.123", false),
+
+            // Trailing period doesn't always match
+            ("foo.com", "example.bar.", false),
+            ("foo.com", "foo.www.", false),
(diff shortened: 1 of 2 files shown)


Tests/X509Tests/DNSNamesTests.swift
```diff
@@ -232,6 +232,10 @@ final class DNSNamesTests: XCTestCase {
             ("1234567.example.com", "example.com", true),
             ("foo.1234567.example.com", "foo.1234567.example.com", true),
             ("foo.example.123", "foo.example.123", false),
+
+            // Trailing period doesn't always match
+            ("foo.com", "example.bar.", false),
+            ("foo.com", "foo.www.", false),
```
dnadoba, line 238: Should we separate our custom test cases from the webpki cases to make it easier if we ever want to update the tests if webkpi adds new cases?

dnadoba (approved):
LGTM

Tests/X509Tests/DNSNamesTests.swift
Lukasa: I don't think it's really necessary, there are already custom cases in there, but they're strictly at the bottom. I also had to tweak a few of the WebPKI test cases because they're testing a slightly different model, so we've already diverged.

PR #42, opened 2023-03-16
Add support for name constraints on URIs

Lukasa:
This is the last of the major constraint types we intend to support initially in name constraints. Mostly URIs leverage the domain name support, but we do have a need to parse URIs in order to extract their host component.


dnadoba (approved):
LGTM

PR #45, opened 2023-03-17
Produce functional NameConstraints

Lukasa:
This patch ties together the various checks we have already written into a single NameConstraintsPolicy, and attaches it to the RFC5280Policy. This includes defining the logic for traversing names, as well as for managing the excluded and permitted subtrees.

Unfortunately, this patch is giant. I wish I could do more about it, but fundamentally the majority of the code it adds is tests, and those tests are all necessary.

### DIFF
--- a/Sources/X509/Verifier/RFC5280/NameConstraintsPolicy.swift
+++ b/Sources/X509/Verifier/RFC5280/NameConstraintsPolicy.swift
@@ -21,7 +21,224 @@ struct NameConstraintsPolicy: VerifierPolicy {
 
     @inlinable
     func chainMeetsPolicyRequirements(chain: UnverifiedCertificateChain) -> PolicyEvaluationResult {
-        // For now we don't implement this, we'll add support later.
+        // The rules for name constraints come from https://www.rfc-editor.org/rfc/rfc5280#section-4.2.1.10.
+        //
+        // Some notes:
+        //
+        // - RFC 5280 says we MUST validate directoryName constraints, and SHOULD validate rfc822Name,
+        //       URI, dNSName, and iPAddress constraints.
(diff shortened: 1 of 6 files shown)


Sources/X509/Verifier/RFC5280/NameConstraintsPolicy.swift
```diff
@@ -21,7 +21,224 @@ struct NameConstraintsPolicy: VerifierPolicy {
...
+        // Our algorithm is recursive: starting from the root and moving towards the leaf, for each CA
+        // cert we apply the name constraints to all of the other certificates in the chain. The one exception
+        // is for self-signed certs where, much like with basic constraints, we briefly pretend that the
+        // self-signed cert issued itself and enforce its own name constraints on it.
+        if chain.count == 1 {
+            return Self._validateNameConstraints(chain[...], issuer: chain.first!)
+        }
+
+        var issuedCerts = chain[...]
+        while let issuer = issuedCerts.popLast(), issuedCerts.count > 0 {
```
dnadoba, line 41: Took me a hot second to understand this. Clever and performant use of `ArraySlice` though.

dnadoba (approved):
LGTM

Side note unrelated to this PR:
`GeneralName.dNSName` and `GeneralName.iPAddress` don't follow the [Swift API Design guidelines case convention](https://www.swift.org/documentation/api-design-guidelines/#follow-case-conventions). 
They should be `.dnsName` and `.ipAddress` even. It doesn't match the ASN.1 capitalization but I think that is fine. Can we still change that? Happy to do it if we agree on it.

Lukasa:
Happy to change the names to match the API design convention. Want to open a PR to do it?

dnadoba:
Yes, will do

PR #47, opened 2023-03-22
Add support for policing critical extensions

Lukasa:
We are required to ensure that we do not validate chains that contain certificates with extensions marked critical that we do not understand. Because we allow pluggable policies which can handle arbitrary extensions, we need a way to ask verifier policy objects what extensions they can handle.

This patch adds support for policies to express that set, as well as for the verifier to ensure that certificates that have critical extensions we cannot understand are not used for chain building. This is integrated early into chain building to reject unsatisfiable chains, and works transparently.

This PR also updates the tests to meet this new standard, including applying the RFC 5280 policy liberally throughout the test suite.

### DIFF
--- a/Sources/X509/Verifier/VerifierPolicy.swift
+++ b/Sources/X509/Verifier/VerifierPolicy.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftCertificates open source project
 //
-// Copyright (c) 2022 Apple Inc. and the SwiftCertificates project authors
+// Copyright (c) 2022-2023 Apple Inc. and the SwiftCertificates project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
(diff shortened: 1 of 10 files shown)


Sources/X509/Verifier/VerifierPolicy.swift
```diff
@@ -11,8 +11,45 @@
...
+    /// X.509 certificates can have extensions marked as `critical`. These extensions _must_ be understood and enforced by the
+    /// verifier. If they aren't understood or processed, then verifying the chain must fail.
+    ///
+    /// ``Verifier`` uses the ``VerifierPolicy/processedExtensions`` field to determine what extensions are understood by a given
+    /// ``PolicySet``. A ``PolicySet`` understands the union of all the understood extensions of its contained ``VerifierPolicy``
+    /// objects.
+    ///
+    /// This may be an empty array, if the policy does not concern itself with any particular extensions. Users must only put
+    /// an extension value in this space if they are actually enforcing the rules of that particular extension value.
+    var processedExtensions: [ASN1ObjectIdentifier] { get }
```
dnadoba, line 43: Shouldn't this be a `Set<ASN1ObjectIdentifier>`?

Sources/X509/Verifier/VerifierPolicy.swift
```diff
@@ -27,11 +64,22 @@ public enum PolicyEvaluationResult: Sendable {
...
         self.policies = policies
+
+        var extensions: [ASN1ObjectIdentifier] = []
+        extensions.reserveCapacity(policies.reduce(into: 0, { $0 += $1.processedExtensions.count }))
+
+        for policy in policies {
+            extensions.append(contentsOf: policy.processedExtensions)
+        }
+
+        self.processedExtensions = extensions
```
dnadoba, line 82: hm... we compute that upfront and not on demand. Therefore a `VerifierPolicy` can never check any policies dynamically e.g. it can't verify an extension only if another extensions is present as well. Is this on purpose? If it is on purpose, I think we should rename this property to reflect that this is not about the extensions it has processed but which it promises to process

Sources/X509/Verifier/VerifierPolicy.swift
Lukasa: I was trying to avoid the overhead of `Hashable` here. Having duplicates really doesn't matter, and it saves us some complexity cost.

Sources/X509/Verifier/VerifierPolicy.swift
dnadoba: If we don't make the list a `Set` we could also just make it a function
```swift
protocol VerifierPolicy {
    func willVerifyCiritalExtension(_ extension: ASN1ObjectIdentifier) -> Bool
}
```
If we want to keep the property we can call it `var verifyingCriticalExtensions: [ASN1ObjectIdentiifer]` 

I kinda want to make it a static func/property but this will only work if `PolicySet` is a variadic generic type.

dnadoba (approved):

PR #59, opened 2023-04-11
Certificate Signing Request

Lukasa:
This patch adds support for working with Certificate Signing Requests to
swift-certificates.

CSRs are a common feature of X.509 based PKIs. They're used to encapsulate
a number of details about the certificate being requested, and are
typically handed to CAs in exchange for a certificate. Their purpose is
to prove possession of a private key, and optionally some other useful
pieces of data.

This patch provides a complete representation of the CSR type, as well as
semantic types to cover the attributes that can be applied to a CSR.
Additionally, CSR signature validation is also available.


dnadoba (changes requested):
Looks mostly good, see inline comments for a couple things I think we should change before merging.

Sources/X509/CSR/CSRAttributes.swift
```diff
@@ -0,0 +1,153 @@
...
+    ///
+    /// - Parameter oid: The OID to search for.
+    @inlinable
+    public subscript(oid oid: ASN1ObjectIdentifier) -> CertificateSigningRequest.Attribute? {
+        get {
+            return self.first(where: { $0.oid == oid })
+        }
+        set {
+            if let newValue = newValue {
+                precondition(oid == newValue.oid)
```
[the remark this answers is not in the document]
Lukasa: That feels fairly un-handy, and prevents us avoiding CoW in some cases.

Sources/X509/CSR/CSRAttribute.swift
```diff
@@ -47,7 +47,34 @@ extension CertificateSigningRequest {
...
+    public func hash(into hasher: inout Hasher) {
+        hasher.combine(self.oid)
+
+        // This achieves order-independent hashing without
+        // having to sort anything.
+        var hash = 0
+        for element in self.values {
+            var newHasher = Hasher()
+            element.hash(into: &newHasher)
+            hash ^= newHasher.finalize()
```
dnadoba, line 72: Nit: the XOR trick is neat but can we add a comment which explains why this does achieve order-independent hashing? E.g. something that says that any associative operation would do it?

dnadoba (commented):
Looks great minus one nit

Sources/X509/CSR/CSRAttribute.swift
```diff
@@ -0,0 +1,122 @@
...
+        }
+    }
+}
+
+extension CertificateSigningRequest.Attribute: Hashable {
+    @inlinable
+    public static func ==(lhs: CertificateSigningRequest.Attribute, rhs: CertificateSigningRequest.Attribute) -> Bool {
+        if lhs.oid != rhs.oid { return false }
+
+        for element in lhs.values {
```
dnadoba, line 55: `lhs.values.count == rhs.values.count` must be true as well or they will be equal even if `rhs` has additional elements.
Can we add for it as well?

Same needs to be done for `CertificateSigningRequest.Attributes`

dnadoba (changes requested):
Catched something on a second look

Sources/X509/CSR/CSRAttribute.swift
Lukasa: Oh good catch!

Lukasa:
Addressed those comments, thanks!

dnadoba (approved):
LGTM

PR #61, opened 2023-04-12
OCSP soft failure mode

dnadoba:
Soft failure mode allows to only trust OCSP responses once the responders certificate is fully verified. If the OCSP fails at any other step e.g. the request could not complete or decoding failed we still meet the OCSP policy in soft failure mode. This should generally be the default unless otherwise required.

### DIFF
--- a/Sources/X509/OCSP/OCSPPolicy.swift
+++ b/Sources/X509/OCSP/OCSPPolicy.swift
@@ -112,6 +112,9 @@ enum OCSPRequestHashAlgorithm {
     }
 }
 
+
+/// Defines the behaviour of ``OCSPVerifierPolicy`` in the event of a failure.
+/// ``soft`` should be used most of the time and will only fail verification if a verified OCSP response reports a status of revoked.
 public struct OCSPFailureMode: Hashable, Sendable {
     /// ``soft`` failure mode will only fail verification if a verified and valid OCSP response reports a status of revoked.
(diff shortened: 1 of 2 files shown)


Sources/X509/OCSP/OCSPPolicy.swift
```diff
@@ -112,6 +112,23 @@ enum OCSPRequestHashAlgorithm {
     }
 }
 
+public struct OCSPFailureMode: Hashable, Sendable {
```
Lukasa, line 115: Can you add a doc comment up here too?

Sources/X509/OCSP/OCSPPolicy.swift
```diff
@@ -131,7 +148,11 @@ public struct OCSPVerifierPolicy<Requester: OCSPRequester>: VerifierPolicy {
     /// currently only set to false for testing
     var nonceExtensionEnabled: Bool = true
     
-    public init(requester: Requester, validationTime: Date) {
+    private var failureMode: OCSPFailureMode
+    
+    
+    public init(mode: OCSPFailureMode, requester: Requester, validationTime: Date) {
```
Lukasa, line 154:
```suggestion
    public init(failureMode: OCSPFailureMode, requester: Requester, validationTime: Date) {
```

dnadoba:

PR #64, opened 2023-04-13
Add API to construct a `SubjectKeyIdentifier` by hashing a `Certificate.PublicKey`

dnadoba:
Adds an initialiser to `SubjectKeyIdentifier` that hashes a given `Certificate.PublicKey` with SHA-1 according to RFC 5280 Section 4.2.1.2.

### DIFF
--- a/Sources/X509/Extension Types/SubjectKeyIdentifier.swift
+++ b/Sources/X509/Extension Types/SubjectKeyIdentifier.swift
@@ -13,6 +13,8 @@
 //===----------------------------------------------------------------------===//
 
 import SwiftASN1
+import Crypto
+import struct Foundation.Data
 
 /// Provides a means of identifying a certificate that contains a particular public key.
 ///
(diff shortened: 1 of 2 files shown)


Sources/X509/Extension Types/SubjectKeyIdentifier.swift
```diff
@@ -78,3 +80,17 @@ extension SubjectKeyIdentifier: CertificateExtensionConvertible {
         return try .init(self, critical: false)
     }
 }
+
+extension SubjectKeyIdentifier {
+    /// Construct a ``SubjectKeyIdentifier`` by hashing the given `publicKey` with SHA-1 according to RFC 5280 Section 4.2.1.2.
+    /// - Parameter publicKey: the public key which will be hashed
+    @inlinable
+    public init(hash publicKey: Certificate.PublicKey) {
```
artemredkin, line 88: I wonder if it's worth adding a shortcut, similar to BoringSSL, where you can just specify `"hash"` and it will use public key from the certificate itself, wdyt?

Lukasa (approved):

Sources/X509/Extension Types/SubjectKeyIdentifier.swift
Lukasa: I think I'm broadly opposed to it. It's really quite hard to construct an API where that functions sensibly.

PR #65, opened 2023-04-13
Expose a CMSSignature type

Lukasa:
There are some use-cases where interrogating the structure of a CMS signature is necessary. This patch adds an SPI for looking at a CMS signature.

As with the rest of our CMS operations, we're not willing to commit to an API shape yet, so this patch doesn't provide one. Instead, users need to opt-in to using the unstable API.


dnadoba (approved):
LGTM

PR #67, opened 2023-04-18
Exclude Secure Enclave related code from CMake build

yim-lee:
Motivation:
macOS CMake build is failing with error:

```
/Users/ec2-user/jenkins/workspace/swift-package-manager-PR-macos-smoke-test/branch-main/swift-certificates/Sources/X509/CertificatePrivateKey.swift:158:32: error: cannot find type 'SecureEnclave' in scope
        case secureEnclaveP256(SecureEnclave.P256.Signing.PrivateKey)
                               ^~~~~~~~~~~~~
ninja: build stopped: subcommand failed.
```

https://ci.swift.org/job/swift-package-manager-PR-macos-smoke-test/2207/console

Modifications:
Introduce compiler directive `INCLUDE_SECURE_ENCLAVE_SUPPORT` to control when to include Secure Enclave related code. It is set in SwiftPM build (`Package.swift`) but not CMake.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,6 +37,9 @@ if(BUILD_SHARED_LIBS)
   set(CMAKE_POSITION_INDEPENDENT_CODE YES)
 endif()
 
+## Uncomment line below to include code that provides Secure Enclave support
+##add_compile_definitions(INCLUDE_SECURE_ENCLAVE_SUPPORT)
+
 if(NOT CMAKE_SYSTEM_NAME STREQUAL Darwin)
   find_package(dispatch CONFIG)
   find_package(Foundation CONFIG)
--- a/Package.swift
+++ b/Package.swift
@@ -16,6 +16,11 @@
 import PackageDescription
 import class Foundation.ProcessInfo
 
+var swiftSettings: [SwiftSetting] = []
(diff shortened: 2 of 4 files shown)


tomerd (approved):

yim-lee:
I tried to make the CMake build compile by importing `CryptoKit`, but ran into various method mismatch and type ambiguity errors between `CryptoKit` and `Crypto`. It's also not possible to isolate the Secure Enclave code because it involves enum cases. Since SwiftPM doesn't use that code, I am excluding it from the CMake build for now to unblock https://github.com/apple/swift-package-manager/pull/6316.

CMakeLists.txt
```diff
@@ -37,6 +37,9 @@ if(BUILD_SHARED_LIBS)
   set(CMAKE_POSITION_INDEPENDENT_CODE YES)
 endif()
 
+## Uncomment line below to include code that provides Secure Enclave support
+##add_compile_definitions(INCLUDE_SECURE_ENCLAVE_SUPPORT)
```
yim-lee, line 41: @etcwilde `add_compile_definitions` seems to work but is there a better way to set `INCLUDE_SECURE_ENCLAVE_SUPPORT`?

PR #68, opened 2023-04-18
Proper fix for 'SecureEnclave' issue in cmake build

yim-lee:
Motivation:
https://github.com/apple/swift-certificates/pull/67 introduced a temporary workaround to unblock cmake build on macOS, but we have a proper fix now (https://github.com/apple/swift-crypto/pull/177).

Modifications:
- Revert https://github.com/apple/swift-certificates/pull/67
- Take new version of `swift-crypto` containing https://github.com/apple/swift-crypto/pull/177

### DIFF
--- a/Sources/X509/CertificatePrivateKey.swift
+++ b/Sources/X509/CertificatePrivateKey.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftCertificates open source project
 //
-// Copyright (c) 2022-2023 Apple Inc. and the SwiftCertificates project authors
+// Copyright (c) 2022 Apple Inc. and the SwiftCertificates project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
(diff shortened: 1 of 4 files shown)


yim-lee:
Requires https://github.com/apple/swift-crypto/pull/177

Sources/X509/CertificatePrivateKey.swift
```diff
@@ -63,7 +63,7 @@ extension Certificate {
             self.backing = .rsa(rsa)
         }
         
-        #if INCLUDE_SECURE_ENCLAVE_SUPPORT
+        #if os(macOS) || os(iOS) || os(watchOS) || os(tvOS)
```
tomerd, line 66: can this be `canImport(Darwin)` or too broad?

Lukasa: It can be, but Yim's patch here is just reverting the change back to what it was, which I quite like. We can do a follow-up PR for `#canImport(Darwin)`.

Lukasa (approved):

PR #70, opened 2023-04-19
Fix wrong variable in documentation

LukasReschke:
`name` doesn't exist in this context and probably refers to `subjectName` and `issuerName`


dnadoba (approved):
Good catch, thank you!

dnadoba:

PR #71, opened 2023-04-20
Support PEM serialisation and parsing of `Certificate`, `Certificate.PublicKey` and `Certificate.PrivateKey`

dnadoba:
### Motivation
PEM is a common format and we should have first class support for it.

### Changes
- require `swift-asn1` `0.8.0`
- conform `Certificate` to `PEMRepresentable`
- add similar methods from `PEMRepresentable` to  `Certificate.PublicKey` and `Certificate.PrivateKey` but don't conform to `PEMRepresentable`. They can't because the are not and should be `DERParseable/Serializable`

### Result
`Certificate`, `Certificate.PublicKey` and `Certificate.PrivateKey` can now be serialised to PEM and parsed from PEM.


Sources/X509/CertificatePrivateKey.swift
```diff
@@ -202,3 +202,69 @@ extension Certificate.PrivateKey {
...
+    @inlinable
+    init(algorithm: AlgorithmIdentifier?, rawEncodedPrivateKey: ArraySlice<UInt8>) throws {
+        switch algorithm {
+        case .some(.ecdsaP256):
+            self = try .init(P256.Signing.PrivateKey(rawRepresentation: rawEncodedPrivateKey))
+        case .some(.ecdsaP384):
+            self = try .init(P384.Signing.PrivateKey(rawRepresentation: rawEncodedPrivateKey))
+        case .some(.ecdsaP521):
+            self = try .init(P521.Signing.PrivateKey(rawRepresentation: rawEncodedPrivateKey))
+        default:
```
dnadoba, line 252: I think we are missing support for RSA keys here but I couldn't find the algorithm identifier from the RFCs or CryptoKit code.
cc @Lukasa

Lukasa: The algorithm identifier should be `.rsaEncryption`, I believe.

Sources/X509/CertificatePrivateKey.swift
```diff
@@ -202,3 +202,81 @@ extension Certificate.PrivateKey {
...
+    init(ecdsaAlgorithm: AlgorithmIdentifier?, rawEncodedPrivateKey: ArraySlice<UInt8>) throws {
+        switch ecdsaAlgorithm {
+        case .some(.ecdsaP256):
+            self = try .init(P256.Signing.PrivateKey(rawRepresentation: rawEncodedPrivateKey))
+        case .some(.ecdsaP384):
+            self = try .init(P384.Signing.PrivateKey(rawRepresentation: rawEncodedPrivateKey))
+        case .some(.ecdsaP521):
+            self = try .init(P521.Signing.PrivateKey(rawRepresentation: rawEncodedPrivateKey))
+        default:
+            throw CertificateError.unsupportedPrivateKey(reason: "unknown algorithm \(ecdsaAlgorithm as Any)")
```
Lukasa, line 265: Same note on `as Any` here.

Lukasa: I generally prefer `String(describing:)` as the fixit suggests

dnadoba: oh, good idea. I somehow though it suggested `as Any` but this is probably just me remembering an old warning and not reading the new message 😅 Thanks!

PR #72, opened 2023-04-21
X509: make the code more portable

compnerd:
`CoreFoundation` is not available on all platforms.  It never was meant to be available on Linux, but was accidentally pushed into the release and is now stuck.  Windows intentionally did not make this module visible.  With the future work with the Foundation rewrite `CoreFoundation` will disappear in more locations.

Windows does not support `inet_pton` except unless the string is explicitly ANSI and not UTF-8.  We must use the proper spelling of `InetPtonW` which takes a UTF-16 string.

The import of `CoreFoundation` was more for the dependency on the C library as a side-effect.  On Windows, the networking functions are provided by a separate user space library rather than the C library. This is now highlighted in the imports.

### DIFF
--- a/Sources/X509/Verifier/RFC5280/URIConstraints.swift
+++ b/Sources/X509/Verifier/RFC5280/URIConstraints.swift
@@ -12,8 +12,14 @@
 //
 //===----------------------------------------------------------------------===//
 
-import CoreFoundation
 import Foundation
+#if os(Windows)
+import WinSDK
+#elseif os(Linux) || os(Android)
+import Glibc
+#elseif os(iOS) || os(macOS) || os(tvOS) || os(watchOS)
+import Darwin
+#endif
 
 extension NameConstraintsPolicy {
     /// Validates that a URI name matches a name constraint.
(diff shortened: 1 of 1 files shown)


dnadoba (approved):
LGTM. Thanks!

Lukasa:

Lukasa (approved):
Thanks for this @compnerd!

Lukasa:
```
18:12:57 /code/Sources/X509/Verifier/RFC5280/URIConstraints.swift:76:24: error: initializer 'init()' cannot be used in an '@inlinable' function because 'CoreFoundation' was not imported by this file; this is an error in Swift 6
18:12:57         var ipv4Addr = in_addr()
18:12:57                        ^
18:12:57 /code/Sources/X509/Verifier/RFC5280/URIConstraints.swift:76:24: note: The missing import of module 'CoreFoundation' will be added implicitly
18:12:57         var ipv4Addr = in_addr()
18:12:57                        ^
```

I don't think "Core Foundation isn't needed" is _quite_ right. Re-adding it on Linux.

Lukasa:

Lukasa:
Transient failure on 5.8 that we have seen before.

Lukasa:
Rather than re-run, I'm going to merge over that failure.

PR #77, opened 2023-04-26
Fix signature comparison

Lukasa:
This patch fixes up an issue where a signature produced from raw form will have a different in-memory representation from the same one deserialized from the network.

### DIFF
--- a/Sources/X509/X509BaseTypes/ECDSASignature.swift
+++ b/Sources/X509/X509BaseTypes/ECDSASignature.swift
@@ -65,8 +65,8 @@ struct ECDSASignature: DERImplicitlyTaggable, Hashable, Sendable {
     @inlinable
     init(rawSignatureBytes raw: Data) {
         let half = raw.count / 2
-        let r = Array(raw.prefix(upTo: half))[...]
-        let s = Array(raw.suffix(from: half))[...]
+        let r = ArraySlice(normalisingToASN1IntegerForm: raw.prefix(upTo: half))
+        let s = ArraySlice(normalisingToASN1IntegerForm: raw.suffix(from: half))
 
         self = ECDSASignature(r: r, s: s)
     }
@@ -165,5 +165,17 @@ extension P521.Signing.ECDSASignature {
         } catch {
             return nil
         }
+    }
+}
+
+extension ArraySlice where Element == UInt8 {
+    /// Normalizes a sequence of bytes that represent an unsigned big endian raw integer into the
+    /// form we'd get from decoding an ASN1 integer.
+    ///
+    /// This means we strip leading zero bytes.
+    @inlinable
+    init<Bytes: RandomAccessCollection>(normalisingToASN1IntegerForm bigEndianRawInteger: Bytes) where Bytes.Element == UInt8 {
+        let realBytes = bigEndianRawInteger.drop(while: { $0 == 0 })
+        self = ArraySlice(realBytes)
     }
 }
--- a/Sources/X509/CertificateSerialNumber.swift
+++ b/Sources/X509/CertificateSerialNumber.swift
@@ -70,7 +70,7 @@ extension Certificate {
         @inlinable
         internal init(generator: inout some RandomNumberGenerator) {
             // drop leading zeros as required by the ASN.1 spec for INTEGERs
-            self.bytes = generator.bytes(count: 20).drop(while: { $0 == 0 })
+            self.bytes = ArraySlice(normalisingToASN1IntegerForm: generator.bytes(count: 20))
         }
     }
 }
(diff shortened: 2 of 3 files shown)


Sources/X509/X509BaseTypes/ECDSASignature.swift
```diff
@@ -167,3 +167,15 @@ extension P521.Signing.ECDSASignature {
...
 }
+
+extension ArraySlice where Element == UInt8 {
+    /// Normalizes a sequence of bytes that represent an unsigned big endian raw integer into the
+    /// form we'd get from decoding an ASN1 integer.
+    ///
+    /// This means we strip leading zero bytes.
+    @inlinable
+    init<Bytes: RandomAccessCollection>(normalisingToASN1IntegerForm bigEndianRawInteger: Bytes) where Bytes.Element == UInt8 {
+        let realBytes = bigEndianRawInteger.drop(while: { $0 == 0 })
```
dnadoba, line 178: There is a bug in here for `Certificate.SerialNumber` which always needs to be positive:

>    CAs MUST force the serialNumber to be a non-negative integer, that
>    is, the sign bit in the DER encoding of the INTEGER value MUST be
>    zero.  This can be done by adding a leading (leftmost) `00'H octet if
>    necessary.  This removes a potential ambiguity in mapping between a
>    string of octets and an integer value.
> https://datatracker.ietf.org/doc/html/rfc5280#appendix-B


This means we need to add a prefix zero byte iff the top bit of the second byte is `1`. We need to be careful though to not create serial numbers longer than 20 bytes:

> As noted in [Section 5.2.3](https://datatracker.ietf.org/doc/html/rfc5280#section-5.2.3), CRL numbers can be expected to contain
>    long integers.  CRL validators MUST be able to handle cRLNumber
>    values up to 20 octets in length.  Conforming CRL issuers MUST NOT
>    use cRLNumber values longer than 20 octets.
> https://datatracker.ietf.org/doc/html/rfc5280#appendix-B

---

I could not find a requirement for Elliptic Curve Digital Signature Algorithm INTEGERs to be positive in [RFC 3279](https://www.ietf.org/rfc/rfc3279.txt).

If they can be negative, we have an additional requirement: The encoding must be the shortest possible which for negative numbers . This means that if the first byte is all ones and the top bit of the second byte is set it one as well, we need to remove the first all ones byte.

dnadoba (changes requested):
I think it is time to add an `ASN1Integer` type to `swift-asn1` which can also be constructed from a `StaticBigInt`. But this can be done separately.

Sources/X509/X509BaseTypes/ECDSASignature.swift
Lukasa: There is no bug here: this perfectly matches the way the ASN1 integer decode handles things. Remember, the ASN1 code does _not_ produce an ArraySlice<UInt8> that perfectly matches the encoded representation. Instead, it strips the leading zero byte:

https://github.com/apple/swift-asn1/blob/c43fef6dc0192ecf27c2fdef0f5957283adab337/Sources/SwiftASN1/Basic%20ASN1%20Types/ASN1Integer.swift#L75-L81

This code therefore does the same thing, stripping _all_ the zero bytes.

Lukasa: Additionally, `r` and `s` for ECDSA are always positive. This is not a property of their ASN.1 representation but a property of their underlying algorithmic derivation.

dnadoba: Oh, so the serialisation path adds the zero back here: https://github.com/apple/swift-asn1/blob/c43fef6dc0192ecf27c2fdef0f5957283adab337/Sources/SwiftASN1/Basic%20ASN1%20Types/ASN1Integer.swift#L98
Then I still think we have one subtle bug here because this will result in Certificate.SerialNumber with 21 octets and this is not allowed:

> As noted in [Section 5.2.3](https://datatracker.ietf.org/doc/html/rfc5280#section-5.2.3), CRL numbers can be expected to contain
> long integers. CRL validators MUST be able to handle cRLNumber
> values up to 20 octets in length. Conforming CRL issuers MUST NOT
> use cRLNumber values longer than 20 octets.
> https://datatracker.ietf.org/doc/html/rfc5280#appendix-B

Or does the leading zero not count?

This is an old bug and not related to the changes you have done though

dnadoba (approved):

Sources/X509/X509BaseTypes/ECDSASignature.swift
Lukasa: Yeah, the initial leading zero does count, we should fix that bug.

PR #79, opened 2023-05-03
Policy Builder

dnadoba:
`PolicySet` is currently a wrapper over `[any VerifierPolicy]`. In the future we want to adopt variadic generic for this type. To allow us to adopt this without breaking API we can use a `@resultBuilder` to hide the construction of the underlying concrete types so we can change it later. In the future we can then also make the types public if we want to.

The new syntax is also more concise. One example from our test can now be written as the following:
```swift
@PolicyBuilder
func create(_ validationTime: Date) -> some VerifierPolicy {
    switch self {
    case .rfc5280:
        RFC5280Policy(validationTime: validationTime)

    case .expiry:
        ExpiryPolicy(validationTime: validationTime)
        CatchAllPolicy()

    case .basicConstraints:
        BasicConstraintsPolicy()
        CatchAllPolicy()

    case .nameConstraints:
        NameConstraintsPolicy()
        CatchAllPolicy()

    }
}
```
instead of wrapping everything in `PolicySet` like before:
```swift
func create(_ validationTime: Date) -> VerifierPolicy {
    switch self {
    case .rfc5280:
        return RFC5280Policy(validationTime: validationTime)
    case .expiry:
        return PolicySet(policies: [
            ExpiryPolicy(validationTime: validationTime),
            CatchAllPolicy()
        ])
    case .basicConstraints:
        return PolicySet(policies: [
            BasicConstraintsPolicy(),
            CatchAllPolicy()
        ])
    case .nameConstraints:
        return PolicySet(policies: [
            NameConstraintsPolicy(),
            CatchAllPolicy()
        ])
    }
}
```


Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version: 5.7
+// swift-tools-version: 5.9
```
Lukasa, line 1: This is an unfortunate change, do we really need it?

dnadoba: ups, reverted.

Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
```diff
@@ -151,7 +152,7 @@ public enum CMS {
             var untrustedIntermediates = CertificateStore(signedData.certificates!)
             untrustedIntermediates.insert(contentsOf: additionalIntermediateCertificates)
             
-            var verifier = Verifier(rootCertificates: trustRoots, policy: policy)
+            var verifier = Verifier(rootCertificates: trustRoots) { policy }
```
Lukasa, line 155: Can we not pass `policy` directly as an argument rather than invoking it ahead of time?

Sources/X509/Verifier/PolicyBuilder.swift
```diff
@@ -0,0 +1,156 @@
...
+        
+        @inlinable
+        init(first: First, second: Second) {
+            self.first = first
+            self.second = second
+        }
+        
+        @inlinable
+        var verifyingCriticalExtensions: [SwiftASN1.ASN1ObjectIdentifier] {
+            first.verifyingCriticalExtensions + second.verifyingCriticalExtensions
```
Lukasa, line 71: This pattern may want us to re-evaluate the way we handle `verifyingCriticalExtensions`. Currently it's assumed that producing this is fairly cheap.

Sources/X509/Verifier/PolicyBuilder.swift
```diff
@@ -0,0 +1,156 @@
...
+    
+    @inlinable
+    public static func buildPartialBlock(accumulated: some VerifierPolicy, next: some VerifierPolicy) -> some VerifierPolicy {
+        Tuple2(first: accumulated, second: next)
+    }
+}
+
+
+// MARK: if
+extension Optional: VerifierPolicy where Wrapped: VerifierPolicy {
```
Lukasa, line 100: I have slightly mixed feelings about having `Optional` conform to `VerifierPolicy`, I wonder if we should have an internal wrapper type instead.

Lukasa:
I also think we want to add a few more tests to validate the behaviour of policybuilder. Let's just confirm it does what we think it does for all the edge cases. Using code coverage to guide this is probably wise.

Sources/X509/Verifier/PolicyBuilder.swift
dnadoba: hm... we can also cache the result automatically in `buildFinalResult` and return a new type which has a stored property for `verifyingCriticalExtensions`. WDYT?

Lukasa: I think that's probably the best way to do that, actually!

Sources/X509/Verifier/AnyPolicy.swift
```diff
@@ -0,0 +1,85 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftCertificates project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import SwiftASN1
+
+public struct AnyPolicy: VerifierPolicy {
```
Lukasa, line 16: Can we add doc comments here?

Lukasa (approved):

PR #80, opened 2023-05-04
Gracefully fail `Extension` serialization

dnadoba:
### Motivation
We currently crash if `Extension` serialization fails but we should throw an error instead.

### Changes
All `@resultBuilder` `build*` methods do not support throwing. We therefore need to delay throwing until the very end. The `@resultBuilder` final result type is now no longer `Certificate.Extensions` but `Result<Certificate.Extensions, any Error>` which allows us to forward the error. 

### Result
We throw the first serialization error instead of crashing. Note that we still serialize all `Extension`s but only preserve the error of the first failure.

### DIFF
--- a/Sources/X509/ExtensionsBuilder.swift
+++ b/Sources/X509/ExtensionsBuilder.swift
@@ -41,39 +41,66 @@
 @resultBuilder
 public struct ExtensionsBuilder {
     @inlinable
-    public static func buildExpression<Extension: CertificateExtensionConvertible>(_ expression: Extension) -> Certificate.Extensions {
-        // TODO: we really need a way to avoid having this be try!.
-        try! Certificate.Extensions([expression.makeCertificateExtension()])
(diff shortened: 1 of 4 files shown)


Sources/X509/ExtensionsBuilder.swift
```diff
@@ -41,42 +41,69 @@
...
 
     @inlinable
-    public static func buildLimitedAvailability(_ component: Certificate.Extensions) -> Certificate.Extensions {
+    public static func buildLimitedAvailability(_ component: Result<Certificate.Extensions, any Error>) -> Result<Certificate.Extensions, any Error> {
         component
     }
 }
 
+
+func foo() -> Certificate.Extensions {
```
Lukasa, line 101: This seems likely to be left over from development

Lukasa (changes requested):
This broadly looks really good, thanks for fixing this up! Can we add some tests that exercise the failure code path to confirm this works?

Sources/X509/ExtensionsBuilder.swift
dnadoba: Good catch

PR #82, opened 2023-05-09
Actually validate we can parse the signature.

Lukasa:
We've had a test floating around for a while that was validating that we parsed signatures properly, but the test never confirmed that the parsed signature was valid, only that Crypto accepted it.

This patch improves that situation.


dnadoba (approved):
the more testing, the better

PR #84, opened 2023-05-10
Police uniqueness of `Certificate.Extensions`

dnadoba:
### Motivation
Extensions in `Certificate.Extensions` must have unique OIDs. 

### Modification
- add dependency on `swift-collections` to use `OrderedDictionary`
- use `OrderedDictionary` as the internal storage type for `Certificate.Extensions`
- enforce uniqueness and throw from the various initialisers if duplicates are found
- enforce uniqueness in `append(_:)` and throw if an `Extension` with the same `OID` is found. 
- remove `append(contentsOf:)` as we can't implement it without keeping a copy of the removed elements because we want to restore the original state if we find duplicates. Users should call `append(_:)` in a loop instead or construct a new `Certificate.Extensions` from scratch.
- add new non-throwing `update(_:)` and `update(_:insertingAt)` methods to `Certificate.Extensions` which will replace an extension if an extension with the same OID already exists.

### Result
Extensions are now unique in `Certificate.Extensions`.


Sources/X509/Extensions.swift
```diff
@@ -79,23 +79,22 @@ extension Certificate {
...
         /// Produce a new Extensions container from a collection of ``Certificate/Extension``.
         ///
         /// - Parameter extensions: The base extensions.
+        /// - Throws: if multiple extensions have the same OID
         @inlinable
-        public init<Elements>(_ extensions: Elements) where Elements: Sequence, Elements.Element == Extension {
+        public init<Elements>(_ extensions: Elements) throws where Elements: Sequence, Elements.Element == Extension {
             self._extensions = Array(extensions)
+            
+            var extensionOIDs = Set<ASN1ObjectIdentifier>()
```
Lukasa, line 92: We should probably avoid the transient allocation of a `Set` here.

Sources/X509/Extensions.swift
```diff
@@ -133,48 +132,104 @@ extension Certificate.Extensions: Hashable { }
...
+    ///   - index: The index at which to insert the ``Certificate/Extension``, if it doesn't already exist.
+    ///
+    /// - Returns: A pair `(old, index)`, where `old` is the ``Certificate/Extension`` that was
+    ///    replaced, or `nil`no ``Certificate/Extension`` with same ``Certificate/Extension/oid`` was present, and `index`
+    ///    is the index corresponding to the updated (or inserted) ``Certificate/Extension``.
+    @inlinable
+    @discardableResult
+    public mutating func update(
+        _ extension: Certificate.Extension,
+        insertingAt index: Int
```
Lukasa, line 209: The behaviour of this is a bit weird, do we really want it? We only sometimes respect the index, which seems unlikely to be right. Surely we should _always_ respect the index.

dnadoba: I have mirrored the API from `OrderedDictionary` as this type is essentially a `SmallOrderedDictionary`: https://swiftpackageindex.com/apple/swift-collections/1.0.4/documentation/orderedcollections/ordereddictionary/updatevalue(_:forkey:insertingat:)

Sources/X509/Extensions.swift
Lukasa: Yup, but for small `n` it's vastly cheaper than allocating the Set and doing the hashing.

Sources/X509/Extensions.swift
dnadoba: Thinking about this more I think it would be weird to do always insert it at `index` because we would need to first remove an extension from `self ` with same `oid` which modifies the collection and kind of invalidates the `index`. You can only safely do it if you know beforehand if the element is present or not or otherwise the index might be outside the bounds.

Sources/X509/Extensions.swift
```diff
@@ -133,48 +137,77 @@ extension Certificate.Extensions: Hashable { }
...
+        }
+    }
+    
+    /// Removes the ``Certificate/Extension`` with the given `oid`.
+    /// - Parameter oid: The  ``Certificate/Extension/oid`` of the``Certificate/Extension`` to remove.
+    /// - Returns: The ``Certificate/Extension`` that was removed,
+    ///     or `nil` if an ``Certificate/Extension`` was not present in with the given `oid`.
+    @inlinable
+    @discardableResult
+    public mutating func remove(for oid: ASN1ObjectIdentifier) -> Certificate.Extension? {
```
Lukasa, line 206: I don't think `for` is the right name for this parameter. Probably fine to have the parameter name be `_`.

Sources/X509/Extensions.swift
```diff
@@ -133,48 +137,77 @@ extension Certificate.Extensions: Hashable { }
...
     ///
-    /// - Parameter extensions: The sequence of new ``Certificate/Extension``s to insert.
+    /// - Parameter extension: The ``Certificate/Extension`` to update or append.
+    /// - Returns: The old ``Certificate/Extension`` that was replaced or `nil` if no ``Certificate/Extension`` with same ``Certificate/Extension/oid`` was present
     @inlinable
-    public mutating func append<Extensions: Sequence>(contentsOf extensions: Extensions) where Extensions.Element == Certificate.Extension {
-        // TODO(cory): enforce uniqueness
-        self._extensions.append(contentsOf: extensions)
+    @discardableResult
+    public mutating func update(_ extension: Certificate.Extension) -> Certificate.Extension? {
```
Lukasa, line 189: I don't think we need `update`, the `set` is sufficient.

Lukasa (approved):

PR #85, opened 2023-05-18
Make expiry checking cheaper.

Lukasa:
Currently, certificate verification is serverely bottlenecked on expiry checking. This bottleneck is almost entirely caused by converting Time objects into Date objects via DateComponents.

We can reduce the cost of this by doing fewer transformations. Instead of converting all the Time objects to Dates, we can convert a Date to a GeneralizedTime and then do all comparisons in GeneralizedTime space.

This vastly reduces the cost of cert validation.


dnadoba (approved):
LGTM

PR #86, opened 2023-05-23
Make constructing Dates cheaper

Lukasa:
Translating between Date and Time is expensive when we use DateComponents, because we jump through ICU unnecessarily. We can move over to standard UTC date math instead, which makes this process vastly cheaper.

### DIFF
--- a/Sources/X509/X509BaseTypes/Time.swift
+++ b/Sources/X509/X509BaseTypes/Time.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the SwiftCertificates open source project
 //
-// Copyright (c) 2022 Apple Inc. and the SwiftCertificates project authors
+// Copyright (c) 2022-2023 Apple Inc. and the SwiftCertificates project authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
@@ -47,33 +47,15 @@ enum Time: DERParseable, DERSerializable, Hashable, Sendable {
 
     @inlinable
     static func makeTime(from date: Date) throws -> Time {
-        let components = gregorianCalendar.dateComponents(in: utcTimeZone, from: date)
+        let components = date.utcDate
 
         // The rule is if the year is outside the range 1950-2049 inclusive, we should encode
         // it as a generalized time. Otherwise, use a UTCTime.
-        // These force-unwraps are safe: all the components are returned by the above call.
-        if (1950..<2050).contains(components.year!) {
-            let utcTime = try UTCTime(
-                year: components.year!,
-                month: components.month!,
-                day: components.day!,
-                hours: components.hour!,
-                minutes: components.minute!,
-                seconds: components.second!
-            )
-
+        if ((1950)..<(2050)).contains(components.year) {
+            let utcTime = try UTCTime(components)
             return .utcTime(utcTime)
         } else {
-            let generalizedTime = try GeneralizedTime(
-                year: components.year!,
-                month: components.month!,
-                day: components.day!,
-                hours: components.hour!,
-                minutes: components.minute!,
-                seconds: components.second!,
-                fractionalSeconds: 0.0
-            )
-
+            let generalizedTime = try GeneralizedTime(components)
             return .generalTime(generalizedTime)
         }
     }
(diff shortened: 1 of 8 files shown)


Sources/X509/X509BaseTypes/TimeCalculations.swift
```diff
@@ -0,0 +1,263 @@
...
+        remdays &-= remyears &* 365
+
+        // Unchecked multiplication here is safe: each of these has earlier been multiplied by
+        // a much larger number and we didn't need checked math then, so we don't need it now.
+        var years = remyears + (4 &* qCycles) + (100 &* cCycles) + (400 &* qcCycles)
+
+        var months = 0
+        while Int64.daysInMonth(months) <= remdays {
+            remdays -= Int64.daysInMonth(months)
+            months += 1
```
dnadoba, line 130: This can be unchecked as well because `Int64.daysInMonth(months)` will fatal error for anything outside `0..<12`
```suggestion
            months &+= 1
```

Sources/X509/X509BaseTypes/TimeCalculations.swift
```diff
@@ -0,0 +1,263 @@
...
+        }
+
+        // Now we normalise the months count back to starting in January.
+        // Safe to do unchecked subtraction here becuase for all numbers 10 or
+        // larger we can subtract by 12.
+        if months >= 10 {
+            months &-= 12
+            years += 1
```
dnadoba, line 138: This can be unchecked as well. The previous year calculations says that it multiplies with number which are much smaller than the previous ones so adding just one will be safe.
```suggestion
            years &+= 1
```

Sources/X509/X509BaseTypes/TimeCalculations.swift
```diff
@@ -0,0 +1,263 @@
...
+
+        // Normalise out the values.
+        //
+        // Safe to do unchecked math on the months as we just checked its value above.
+        // Same for remdays, the loop only terminates if the number is smaller than at most 31.
+        //
+        // Note that, unlike struct tm, we return ordinal month numbers as well as days (i.e. 1 to 12).
+        // This fits us better when working with generalizedtime and friends.
+        return (
+            year: Int(years + 2000),
```
dnadoba, line 149: Same here, I think this can be unchecked as well.
```suggestion
            year: Int(years &+ 2000),
```

Sources/X509/X509BaseTypes/Time.swift
```diff
@@ -47,111 +47,55 @@ enum Time: DERParseable, DERSerializable, Hashable, Sendable {
...
             return .generalTime(generalizedTime)
         }
     }
 }
 
 extension Date {
     @inlinable
-    init?(_ time: Time) {
-        let maybeDate: Date?
+    init(fromUTCDate date: (year: Int, month: Int, day: Int, hours: Int, minutes: Int, seconds: Int)) {
```
dnadoba, line 66: Should we introduce a struct for that type?

Sources/X509/X509BaseTypes/TimeCalculations.swift
```diff
@@ -0,0 +1,263 @@
...
+            month: Int(months &+ 3),
+            day: Int(remdays &+ 1),
+            hours: Int(remsecs / 3600),
+            minutes: Int(remsecs / 60 % 60),
+            seconds: Int(remsecs % 60)
+        )
+    }
+
+    @inlinable
+    init(timestampFromUTCDate date: (year: Int, month: Int, day: Int, hours: Int, minutes: Int, seconds: Int)) {
```
dnadoba, line 159: Can we document the valid value ranges for which we guarantee not to crash and add asserts that check that the inputs are in this range?

Sources/X509/X509BaseTypes/TimeCalculations.swift
```diff
@@ -0,0 +1,263 @@
...
+
+        leaps += (97 * cycles) + (24 * centuries) - (isLeap ? 1 : 0)
+        return (seconds: ((year - 100) * 31536000) + (leaps * 86400) + 946684800 + 86400, isLeap: isLeap)
+    }
+
+    @inlinable
+    static func monthToSeconds(_ month: Int64, isLeap: Bool) -> Int64 {
+        var secondsThroughMonth: Int64
+
+        // musl tolerates out-of-band months: we don't.
```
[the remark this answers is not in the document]
Lukasa: It does, it's policed elsewhere.

Sources/X509/X509BaseTypes/TimeCalculations.swift
Lukasa: I don't think that argument holds: while it was true for each element _individually_, it may not be true for the sum of them, which is why the addition between those values was not unchecked.

Sources/X509/X509BaseTypes/Time.swift
Lukasa: I didn't want to, because we already have two very similar structs: `GeneralizedTime` and `UTCTime`.

Lukasa:
I've updated, please re-review.

dnadoba (approved):
Great performance improvement!

Sources/X509/X509BaseTypes/TimeCalculations.swift
```diff
@@ -0,0 +1,292 @@
...
+    static func daysInMonth(_ month: Int) -> Int64 {
+        // This may seem weird, but these months are indexed from _March_.
+        // Thus, month 0 is March, month 11 is February.
+        switch month {
+        case 0, 2, 4, 5, 7, 9, 10:
+            return 31
+        case 1, 3, 6, 8:
+            return 30
+        case 11:
+            return 29
```
3a4oT, line 99: February usually takes `28` days. I didn't get why it was set to 29 as default.

Sources/X509/X509BaseTypes/TimeCalculations.swift
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    @inlinable
+    static var secondsPerYear: Int64 {
+        .secondsPerDay * .daysPerYear
+    }
+
+    @inlinable
+    static var daysPerYear: Int64 {
+        365
```
3a4oT, line 86: If we count February as 29 days long, shouldn't it be set to 366 days?

Sources/X509/X509BaseTypes/TimeCalculations.swift
Lukasa: The use of 29 here is subtle and weird, but we get away with it because all the calculations are performed on a "year" that begins in March. We calculate the year _before_ we calculate the day of the year, and that year calculation is normalized for leap years. Thus, we can only get to the 29th day of February _if_ we already know we're in a leap year: otherwise we'll be in the first day of the following year, which starts on the 1st of March.

Sources/X509/X509BaseTypes/TimeCalculations.swift
Lukasa: Nope: we account for the leap years in this calculation elsewhere.

Sources/X509/X509BaseTypes/TimeCalculations.swift
3a4oT: Thanks for the response!

PR #88, opened 2023-05-25
Remove the "from parts" Certificate constructor.

Lukasa:
The "from parts" Certificate constructor has existed for a while. As noted in the docstring, it was originally added for testing and validation purposes. However, it's not very useful for that purpose.

For one thing, many of the constitutuent parts are not actually directly constructible. For example, one cannot create a Signature type directly.

For another thing, invalid Certificate handling tests probably want to be able to construct Certificates that are invalid in a number of interesting ways, and so will want to use the DER representation to do so.

As a result, this intializer is not broadly useful: even this library doesn't use it for testing. Let's remove it.

### DIFF
--- a/Sources/X509/Certificate.swift
+++ b/Sources/X509/Certificate.swift
@@ -145,54 +145,6 @@ public struct Certificate {
     @usableFromInline
     internal let signatureAlgorithmBytes: ArraySlice<UInt8>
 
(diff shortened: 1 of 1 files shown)


dnadoba (approved):

Lukasa:
Merging over API breakage, which is expected.

PR #89, opened 2023-06-01
Remove unnecessary TODO

Lukasa:
This patch removes a TODO requesting validation of an RSA signature, which we cannot do.


dnadoba (approved):
LGTM

PR #90, opened 2023-06-01
Police uniqueness of `ExtendedKeyUsage`

dnadoba:
### Motivation
`ExtendedKeyUsage` is only allowed to each `Usage` at most once. It models an `OrderedSet` and should have a similar API surface.

### Modification
- `ExtendedKeyUsage.init(_:)` now throws an error if a duplicate `Usage` is encountered.
- add `append(_:)` method to insert a `Usage` at the end of set if not already present.
- add `insert(_:at:)` method to insert a `Usage` at a specific index if not already present.
- add `remove(_:)` method to remove a `Usage` if present.
- remove `subscribe(position:)` setter because it can't safely enforce the uniqueness constraint without crashing. `OrderedSet` doesn't either.

Unrelated to `ExtendedKeyUsage` but related to a similar TODO:
- We don't need to maintain uniqueness for `AuthorityInformationAccess`. Was probably a copy and past error. TODO therefore removed.

### Result
`ExtendedKeyUsage` now properly maintains uniqueness.

### DIFF
--- a/Sources/X509/Extension Types/ExtendedKeyUsage.swift
+++ b/Sources/X509/Extension Types/ExtendedKeyUsage.swift
@@ -26,8 +26,22 @@ public struct ExtendedKeyUsage {
     ///
     /// - Parameter usages: The purposes for which the certificate may be used.
     @inlinable
-    public init<Usages: Sequence>(_ usages: Usages) where Usages.Element == Usage {
+    public init<Usages: Sequence>(_ usages: Usages) throws where Usages.Element == Usage {
         self.usages = Array(usages)
+        
+        // check for duplicates using a linear scan
+        // this is more performant compared to hashing if we have less than ~64 usages
+        for index in self.usages.indices {
+            let usage = self.usages[index]
+            for currentIndex in self.usages.indices.dropFirst(index + 1) {
+                let currentUsage = self.usages[currentIndex]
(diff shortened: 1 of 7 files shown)


Sources/X509/Extension Types/ExtendedKeyUsage.swift
```diff
@@ -26,8 +26,20 @@ public struct ExtendedKeyUsage {
...
+        // this is more performant compared to hashing if we have less than ~64 usages
+        for index in self.usages.indices {
+            let usage = self.usages[index]
+            for currentIndex in self.usages.indices.dropFirst(index + 1) {
+                let currentUsage = self.usages[currentIndex]
+                if usage == currentUsage {
+                    throw CertificateError.duplicateOID(reason: "duplicate \(usage) usage. First at \(index) and second at \(currentIndex)")
+                }
+            }
+        }
```
Lukasa, line 42: We've written this code twice: can you factor the algorithm out to a common helper function and then call it in the relevant places?

dnadoba: Done in https://github.com/apple/swift-certificates/pull/90/commits/cdf557e0cd72defa7b62cb6798dfe1ad3334e0fc

Lukasa (approved):

PR #92, opened 2023-06-02
Implement remove for `RelativeDistinguishedName`

dnadoba:
and remove unnecessary TODO as discussed in https://github.com/apple/swift-certificates/pull/91#issuecomment-1573871378


Sources/X509/RelativeDistinguishedName.swift
```diff
@@ -35,18 +35,13 @@ import SwiftASN1
 public struct RelativeDistinguishedName {
     // TODO: Should we special-case this to the circumstance where we have only one attribute?
     @usableFromInline
-    var attributes: [Attribute] {
-        didSet {
-            Self._sortElements(&self.attributes)
-        }
-    }
```
Lukasa, line 42: Is it appropriate to drop the sort here?

Lukasa (approved):

PR #94, opened 2023-06-12
Implement property setters of `NameConstraints` 

dnadoba:
### Motivation
We want to support fine grained mutations of `NameConstraints`. In addition, the properties of it should only be views into the underlying storage to eliminate allocations costs for inspection `NameConstraints` properties.

### Modification
- properties are now lazy views into the underlying storage
- setters for all properties are implemented

### Result
Mutating `NameConstraints` is now possible and getting the properties is lazy.

### DIFF
--- a/Sources/X509/Extension Types/NameConstraints.swift
+++ b/Sources/X509/Extension Types/NameConstraints.swift
@@ -253,7 +253,7 @@ public struct NameConstraints {
     ///
     /// These restrictions are expressed in forms like `host.example.com`. Any DNS name that can be
     /// constructed by adding zero or more labels to the left-hand side of the name satifies the constraint.
-    public var permittedDNSDomains: DNSNames {
+    public internal(set) var permittedDNSDomains: DNSNames {
         get {
             DNSNames(subtrees: permittedSubtrees)
         }
@@ -272,7 +272,7 @@ public struct NameConstraints {
     ///
     /// These restrictions are expressed in forms like `host.example.com`. Any DNS name that can be
     /// constructed by adding zero or more labels to the left-hand side of the name satifies the constraint.
-    public var excludedDNSDomains: DNSNames {
+    public internal(set) var excludedDNSDomains: DNSNames {
         get {
             DNSNames(subtrees: excludedSubtrees)
         }
@@ -297,7 +297,7 @@ public struct NameConstraints {
     /// This represents a subnet root and its mask.
     ///
     /// Any IP address attested that falls within one of these subnets matches the constraint.
-    public var permittedIPRanges: IPRanges {
+    public internal(set) var permittedIPRanges: IPRanges {
         get {
             IPRanges(subtrees: permittedSubtrees)
         }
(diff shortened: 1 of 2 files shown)


Sources/X509/Extension Types/NameConstraints.swift
```diff
@@ -24,43 +24,266 @@ import SwiftASN1
...
+        
+        @inlinable
+        public func hash(into hasher: inout Hasher) {
+            hasher.combine(contentsOf: self)
+        }
+        
+        @inlinable
+        public var description: String {
+            "[\(self.joined(separator: ", "))]"
+        }
```
Lukasa, line 41: Can we move these down below the stored properties and initializers?

Sources/X509/Extension Types/NameConstraints.swift
```diff
@@ -24,43 +24,266 @@ import SwiftASN1
...
+        
+        @inlinable
+        public func makeIterator() -> some IteratorProtocol<String> {
+            self.subtrees.lazy.compactMap {
                 if case .dnsName(let name) = $0 {
                     return name
                 } else {
                     return nil
                 }
+            }.makeIterator()
```
Lukasa, line 69: This is likely to perform extremely poorly: it should be sufficient to implement `Collection` over an opaque index type and use the free iterator.

Sources/X509/Extension Types/NameConstraints.swift
```diff
@@ -24,43 +24,266 @@ import SwiftASN1
...
         }
         set {
-            fatalError("TODO")
+            excludedSubtrees.removeAll {
+                guard case .dnsName = $0 else {
+                    return false
+                }
+                return true
+            }
+            excludedSubtrees.append(contentsOf: newValue.filtered)
```
Lukasa, line 286: Hmm, these are fairly expensive. I wonder if we can just decide we don't need them.

Sources/X509/Extension Types/NameConstraints.swift
dnadoba: Why do you think this is expensive?

Sources/X509/Extension Types/NameConstraints.swift
dnadoba: I don't think this is too expensive. Why do you think so? `removeAll(where:)` has linear time complexity as well as `append(contentsOf:)`. Any idea how to make it cheaper?

Sources/X509/Extension Types/NameConstraints.swift
Lukasa: The `.lazy` transforms are highly general, and their performance is often spotty.

Sources/X509/Extension Types/NameConstraints.swift
Lukasa: Nope, or I would have suggested one ;).

The awkward reality is that a linear time property setter is rarely what you actually want, which is why I'm asking whether we can safely just remove them altogether.

dnadoba: I have made them internal for now because it makes testing quite nice. If we have a real use case we can think about it again.

Sources/X509/Extension Types/NameConstraints.swift
dnadoba: Adopted the `Collection` protocol with an opaque index

Lukasa (approved):

PR #95, opened 2023-06-13
Limit the number elements in `Extensions` and `ExtendedKeyUsage`

dnadoba:
### Motivation
When decoding a certificate we make sure that `ExtendedKeyUsage` and `Certificate.Extensions` only contain unique elements.
In the real world those collections are fairly small and we therefore use a linear scan to check for duplicates and fail decoding if we detect a duplicate.
However, a bad actor could increase the number of elements to an arbitrary number which will increase our decoding time exponentially.

### Modification
Limit the number of elements we accept to 16 before we do the uniqueness check

### Result
We safely fail decoding for unreasonable large number of elements.

Note that if we encounter a real world use case for more elements we can also switch to hash based uniqueness check and support more elements safely.

### DIFF
--- a/Sources/X509/Extensions.swift
+++ b/Sources/X509/Extensions.swift
@@ -89,6 +89,11 @@ extension Certificate {
         public init<Elements>(_ extensions: Elements) throws where Elements: Sequence, Elements.Element == Extension {
             self._extensions = Array(extensions)
             
+            let maxExtensions = 16
+            guard self._extensions.count <= maxExtensions else {
+                throw ASN1Error.invalidASN1Object(reason: "Too many extensions. Found \(self._extensions.count) but only \(maxExtensions) are allowed.")
+            }
+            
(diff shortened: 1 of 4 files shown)


Sources/X509/Extensions.swift
```diff
@@ -89,6 +89,11 @@ extension Certificate {
         public init<Elements>(_ extensions: Elements) throws where Elements: Sequence, Elements.Element == Extension {
             self._extensions = Array(extensions)
             
+            let maxExtensions = 16
```
Lukasa, line 92: I think we can safely raise this to 32 without hurting us at all, and it's probably safer.

Lukasa:
We should also add a comment above these numbers to clarify that the limit is arbitrary, we're just trying not to get punished :D

Lukasa (approved):

PR #96, opened 2023-06-13
Setup allocation test framework

dnadoba:

PR #98, opened 2023-06-16
Implement TinyArray

dnadoba:
We want to improve allocations for a couple of collections that are only ever expected to store a single element. However, in rare cases they might store multiple elements so they still need to support that.

### DIFF
--- a/IntegrationTests/tests_01_allocation_counters/test_01_resources/test_tiny_array_cow_append_contents_of.swift
+++ b/IntegrationTests/tests_01_allocation_counters/test_01_resources/test_tiny_array_cow_append_contents_of.swift
@@ -12,14 +12,19 @@
 //
 //===----------------------------------------------------------------------===//
 
-@_spi(IntegrationTests) import X509
+import _CertificateInternals
 
 func run(identifier: String) {
     measure(identifier: identifier) {
-        var tinyArray = TinyArray<Int>()
-        for i in 0..<1000 {
-            tinyArray.append(i)
+        var count = 0
+        for _ in 0..<1000 {
+            var tinyArray = _TinyArray<Int>()
+            for i in 0..<1000 {
+                tinyArray.append(i)
+            }
+            count += tinyArray.count
         }
-        return tinyArray.count
+        
+        return count
     }
 }
--- a/IntegrationTests/tests_01_allocation_counters/test_01_resources/test_tiny_array_cow_append_contents_of.swift
+++ b/IntegrationTests/tests_01_allocation_counters/test_01_resources/test_tiny_array_cow_append_contents_of.swift
@@ -16,16 +16,15 @@
 
 func run(identifier: String) {
     measure(identifier: identifier) {
-        var array = Array<Int>()
-        array.reserveCapacity(4)
-        array.append(1)
-        array.append(2)
-        var tinyArray = TinyArray(array)
-        // drop the ref to array so TinyArray is now the single owner
-        array = []
-        
+        var tinyArray = TinyArray<Int>()
+        tinyArray.append(1)
+        tinyArray.append(2)
         tinyArray.append(3)
         tinyArray.append(4)
+        tinyArray.append(5)
+        tinyArray.append(6)
+        tinyArray.append(7)
+        tinyArray.append(8)
         return tinyArray.count
     }
 }
--- a/IntegrationTests/tests_01_allocation_counters/test_01_resources/test_tiny_array_cow_append_contents_of.swift
+++ b/IntegrationTests/tests_01_allocation_counters/test_01_resources/test_tiny_array_cow_append_contents_of.swift
@@ -17,14 +17,9 @@
 func run(identifier: String) {
     measure(identifier: identifier) {
         var tinyArray = TinyArray<Int>()
-        tinyArray.append(1)
-        tinyArray.append(2)
-        tinyArray.append(3)
-        tinyArray.append(4)
-        tinyArray.append(5)
-        tinyArray.append(6)
-        tinyArray.append(7)
-        tinyArray.append(8)
+        for i in 0..<1000 {
+            tinyArray.append(i)
+        }
         return tinyArray.count
     }
 }
--- a/IntegrationTests/tests_01_allocation_counters/test_01_resources/test_tiny_array_cow_append_contents_of.swift
+++ b/IntegrationTests/tests_01_allocation_counters/test_01_resources/test_tiny_array_cow_append_contents_of.swift
@@ -12,10 +12,20 @@
 //
 //===----------------------------------------------------------------------===//
 
-import X509
+@_spi(IntegrationTests) import X509
 
 func run(identifier: String) {
     measure(identifier: identifier) {
-        return 0
+        var array = Array<Int>()
+        array.reserveCapacity(4)
+        array.append(1)
+        array.append(2)
+        var tinyArray = TinyArray(array)
+        // drop the ref to array so TinyArray is now the single owner
+        array = []
+        
+        tinyArray.append(3)
+        tinyArray.append(4)
+        return tinyArray.count
     }
 }
--- a/IntegrationTests/tests_01_allocation_counters/test_01_resources/test_tiny_array_non_allocating_operations.swift
+++ b/IntegrationTests/tests_01_allocation_counters/test_01_resources/test_tiny_array_non_allocating_operations.swift
@@ -15,9 +15,9 @@
 @_spi(IntegrationTests) import X509
 
 func run(identifier: String) {
-    let preallocatedArray1 = [1, 2, 3, 4]
-    let preallocatedArray2 = [1, 2, 3, 4]
-    let preallocatedArray3 = [1, 2, 3, 4]
+    let preallocatedArray1 = [1]
+    let preallocatedArray2 = [1, 2]
+    let preallocatedArray3 = [1, 2, 3]
     let preallocatedArray4 = [1, 2, 3, 4]
 
     measure(identifier: identifier) {
--- a/IntegrationTests/tests_01_allocation_counters/test_01_resources/test_tiny_array_non_allocating_operations.swift
+++ b/IntegrationTests/tests_01_allocation_counters/test_01_resources/test_tiny_array_non_allocating_operations.swift
@@ -15,46 +15,13 @@
 @_spi(IntegrationTests) import X509
 
 func run(identifier: String) {
-    let preallocatedArray1 = [1]
-    let preallocatedArray2 = [1, 2]
-    let preallocatedArray3 = [1, 2, 3]
-    let preallocatedArray4 = [1, 2, 3, 4]
-
     measure(identifier: identifier) {
         var counts = 0
         counts += TinyArray(CollectionOfOne(1)).count
-        counts += TinyArray(preallocatedArray1).count
-        counts += TinyArray(preallocatedArray2).count
-        counts += TinyArray(preallocatedArray3).count
-        counts += TinyArray(preallocatedArray4).count
         
         do {
             var array = TinyArray<Int>()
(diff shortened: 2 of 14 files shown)


Sources/X509/TinyArray.swift
```diff
@@ -0,0 +1,230 @@
...
+                        return
+                    }
+                    elements.reserveCapacity(elements.count + newElements.underestimatedCount)
+                    elements.append(firstElement)
+                    elements.append(secondElement)
+                    elements.appendRemainingElements(from: &iterator)
+                    self = .arbitary(elements)
+                }
+            } else {
+                elements.append(contentsOf: newElements)
```
glbrntt, line 148: I've found the compiler to be very hit and miss with avoiding CoWs in these situations. Might be worth verifying it doesn't CoW or temporarily setting `self = .arbitrary([])`

Sources/X509/TinyArray.swift
```diff
@@ -0,0 +1,230 @@
...
+            guard index == 0 else {
+                fatalError("index \(index) out of bounds")
+            }
+            self = .arbitary([])
+            return oldElement
+            
+        case .arbitary(var elements):
+            defer {
+                if elements.count == 1 {
+                    self = .one(elements[0])
```
glbrntt, line 169: I'm not sure this is a worthwhile optimisation if we're aiming to reduce allocations: if `elements.count == 1` then we _must_ have had some heap storage before. If we drop batch to `.one` and later add another element then we'll be forced to allocate, if we keep the `.arbitrary` storage then we won't allocate when adding the next element. This would of course require you to implement `Hashable`/`Equatable`.

Sources/X509/TinyArray.swift
```diff
@@ -0,0 +1,230 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+
+/// ``TinyArray`` is optimized to store zero or one ``Element``.
+/// It supports arbitrary many elements but if only up to one ``Element`` is stored it does **not** allocate seperate storage on the heap and
+/// instead stores the ``Element`` inline.
+@usableFromInline
+enum TinyArray<Element>: RandomAccessCollection {
```
glbrntt, line 21: I don't love that callers can see it's an `enum` which to me is an implementation detail. That said it's internal so I'm also okay to leave it.

Sources/X509/TinyArray.swift
dnadoba: Could you review https://github.com/apple/swift-certificates/pull/96 as well? Then I can write an allocation test as well

Sources/X509/TinyArray.swift
dnadoba: I have wondered as well if we should do this or not. I have now implemented a custom Equatable/Hashable function and no longer throw the array away and instead store it.

Sources/X509/TinyArray.swift
```diff
@@ -0,0 +1,323 @@
...
+@usableFromInline
+struct TinyArray<Element> {
+    @usableFromInline
+    enum Storage {
+        case one(Element)
+        case arbitrary([Element])
+    }
+    
+    @usableFromInline
+    var storage: Storage
```
Lukasa, line 29: You almost certainly want an `@inlinable init` here.

Sources/X509/TinyArray.swift
```diff
@@ -0,0 +1,323 @@
...
+        switch self {
+        case .one: return 1
+        case .arbitrary(let elements): return elements.endIndex
+        }
+    }
+}
+ 
+extension TinyArray.Storage {
+    @inlinable
+    init(_ elements: some Sequence<Element>) {
```
Lukasa, line 173: This misses an optimization for the case where there is only one element.

Sources/X509/TinyArray.swift
```diff
@@ -0,0 +1,323 @@
...
+            elements.append(firstElement)
+            elements.append(secondElement)
+            elements.appendRemainingElements(from: &iterator)
+            self = .arbitrary(elements)
+            
+        case .arbitrary(var elements):
+            if elements.isEmpty {
+                if let array = newElements as? Array<Element> {
+                    // fast path
+                    self = .arbitrary(array)
```
Lukasa, line 232: Have you measured this fast-path? Runtime type checks are quite expensive.

Sources/X509/TinyArray.swift
dnadoba: They are in line 65 and 70

Sources/X509/TinyArray.swift
dnadoba: `self = .arbitrary([])` doesn't allocate and `self.append(contentsOf:)` has a fast path for a single element.

Sources/X509/TinyArray.swift
dnadoba: If specialised this runtime check is eliminated and evaluated at compile time, otherwise it can be more expensive or less, depending on the size of the array.

dnadoba: Alternatively we can add a specific `append(contentsOf:)` overload for `Array`  but we will not always hit it because `swift-certificates` often only operates on a generic `Sequence` type rather a concrete `Array`. 

For the optimisation I want to use it for we will at the moment never hit this fast-path because we will not have an array in the first place but a non-allocation lazy `Sequence`. Therefore I'm also open to just remove it to get this merged.

Lukasa: I think for now let's just get it merged.

Sources/X509/TinyArray.swift
dnadoba: Added an allocation test in https://github.com/apple/swift-certificates/pull/98/commits/4f1037b5469a63ef89af5c8104690c35fd4c2394 that shows that we only trigger a single allocation and do not trigger CoW on `append(_:)` which calls down to `append(contentsOf:)`

glbrntt (approved):

IntegrationTests/tests_01_allocation_counters/test_01_resources/test_tiny_array_cow_append_contents_of.swift
```diff
@@ -0,0 +1,30 @@
...
+        tinyArray.append(1)
+        tinyArray.append(2)
+        tinyArray.append(3)
+        tinyArray.append(4)
+        tinyArray.append(5)
+        tinyArray.append(6)
+        tinyArray.append(7)
+        tinyArray.append(8)
+        return tinyArray.count
+    }
```
Lukasa, line 29: Typically we try to run the bodies of these in loops that iterate 1000 times, just to distinguish allocations from within the loop from allocations outside of them.

dnadoba: Done in https://github.com/apple/swift-certificates/pull/98/commits/6a42c66519c533f3eb23fa8e6667176d6f60ae4b

IntegrationTests/tests_01_allocation_counters/test_01_resources/test_tiny_array_non_allocating_operations.swift
```diff
@@ -0,0 +1,30 @@
...
+//===----------------------------------------------------------------------===//
+
+@_spi(IntegrationTests) import X509
+
+func run(identifier: String) {
+    measure(identifier: identifier) {
+        var counts = 0
+        counts += TinyArray(CollectionOfOne(1)).count
+        
+        do {
```
Lukasa, line 22: Same note here on needing a loop.

IntegrationTests/tests_01_allocation_counters/test_01_resources/test_tiny_array_cow_append_contents_of.swift
dnadoba: That's what we want to test though. We want to test that we don't trigger CoW so we want to see asymptotical allocation behaviour. If we would trigger CoW we would see ~1000 allocations.

IntegrationTests/tests_01_allocation_counters/test_01_resources/test_tiny_array_non_allocating_operations.swift
Lukasa: Because if you're wrong, you want to tell the difference between "the runtime just happened to fire an allocation from a background thread at this moment" and "the loop body actually allocates deterministically". The easiest way to tell that is to make any allocation that occurs in the loop body happen 1000 times, which we do by wrapping the test in a loop that iterates 1000 times.

IntegrationTests/tests_01_allocation_counters/test_01_resources/test_tiny_array_cow_append_contents_of.swift
Lukasa: Yes, but we want to measure the asymptotic behaviour too. Hence multiplying that out by 1000. We want to know the actual number of allocations in the loop above the noise floor, which we get by multiplying by 1000.

Sources/_CertificateInternals/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+##===----------------------------------------------------------------------===##
+
+add_library(_CertificateInternals
+  "_TinyArray.swift")
+
+set_target_properties(_CertificateInternals PROPERTIES
+  INTERFACE_INCLUDE_DIRECTORIES ${CMAKE_Swift_MODULE_DIRECTORY})
+
+_install_target(_CertificateInternals)
+set_property(GLOBAL APPEND PROPERTY SWIFT_CERTIFICATES_EXPORTS _CertificateInternals)
```
dnadoba, line 22: @yim-lee I'm adding a new internal target with this PR and have copied the make file from the X509 folder. I'm not familiar with cmake. Could you please have a look at this file and the changes in the other cmake related changes?

Sources/_CertificateInternals/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+## See LICENSE.txt for license information
+## See CONTRIBUTORS.txt for the list of SwiftCertificates project authors
+##
+## SPDX-License-Identifier: Apache-2.0
+##
+##===----------------------------------------------------------------------===##
+
+add_library(_CertificateInternals
+  "_TinyArray.swift")
+
```
yim-lee, line 17: Need to add:

```
target_link_libraries(_CertificateInternals PRIVATE
  $<$<NOT:$<PLATFORM_ID:Darwin>>:Foundation>)
```

if the module depends on `Foundation`.

Sources/_CertificateInternals/_TinyArray.swift
```diff
@@ -0,0 +1,315 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftCertificates project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
```
yim-lee, line 15: At quick glance I'm not sure this import is really needed?

Sources/_CertificateInternals/CMakeLists.txt
yim-lee: I left a couple of comments regarding `Foundation` but otherwise cmake changes look good to me.

Sources/_CertificateInternals/_TinyArray.swift
dnadoba: You are right, good catch! I have removed it.

Sources/_CertificateInternals/CMakeLists.txt
dnadoba: Thank you Yim!

PR #100, opened 2023-06-19
Added subjectPublicKeyInfoBytes to public keys

### LINKED ISSUE #99: Get data from public key
I'm trying to use this with vapor to build a micro service capable of handling app attestation. Part of that is:
> Create the SHA256 hash of the public key in credCert, and verify that it matches the key identifier from your app.

Problem is the public key exposes no way to do this. It'd be very helpful if `Certificate.PublicKey` would conform to `DataProtocol`, or expose bytes, or something along those lines.

Tyler-Keith-Thompson:
Closes #99

### DIFF
--- a/Sources/X509/CertificatePublicKey.swift
+++ b/Sources/X509/CertificatePublicKey.swift
@@ -227,6 +227,16 @@ extension SubjectPublicKeyInfo {
     }
 }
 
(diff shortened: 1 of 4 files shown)


dnadoba:
@swift-server-bot please test this

dnadoba:

PR #101, opened 2023-06-21
Do not allocate for `RelativeDistinguishedName`s with just a single attribute

dnadoba:
### Motivation
A `RelativeDistinguishedName` has almost always just a single `Attribute` but we currently always allocate an array.

### Modification
- Use `_TinyArray` instead of `Array` as the internal storage for `Attribute`s in `RelativeDistinguishedName`
- lazily decode nodes. We needed to copy over some code from `swift-asn1`. We might want to think about porting it back to `swift-asn1` but need to think about the right public API for that which we can start in this PR.
- `RelativeDistinguishedName` init can't fail and therefore it is now non-throwing init

### Result
Allocations have improved


Sources/X509/RelativeDistinguishedName.swift
```diff
@@ -179,3 +193,75 @@ extension RelativeDistinguishedName: DERImplicitlyTaggable {
...
+    ///
+    /// Constructs an array of `T` elements parsed from the set.
+    ///
+    /// - parameters:
+    ///     - of: An optional parameter to express the type to decode.
+    ///     - identifier: The ``ASN1Identifier`` that the SET OF is expected to have.
+    ///     - rootNode: The ``ASN1Node`` to parse
+    /// - returns: An array of elements representing the elements in the sequence.
+    @inlinable
+    public static func lazySet<T: DERParseable>(of: T.Type = T.self, identifier: ASN1Identifier, rootNode: ASN1Node) throws -> some Sequence<Result<T, Error>> {
```
Lukasa, line 208: It's pretty unfortunate to have to add this, and its supporting code, here: doubly so that we've made it `public`. Can we push this down into ASN1?

dnadoba: Yes sure, I just didn't like that particular name and couldn't come up with a better one.

Lukasa: I think it's fine.

dnadoba: https://github.com/apple/swift-asn1/pull/32

dnadoba:

dnadoba:
Allocations are reduce by ~7% for parsing certificates

dnadoba:
Looks like the nightly toolchain is still pulling in `swift-asn1` `0.9.1`. Let's see if a rerun fixes it.

dnadoba:

Sources/X509/RelativeDistinguishedName.swift
```diff
@@ -33,23 +34,36 @@ import SwiftASN1
...
+    ///
+    /// - Parameter attributes: The sequence of ``Attribute``s that make up the ``DistinguishedName``.
+    @inlinable
+    public init<AttributeSequence: Sequence>(_ attributes: AttributeSequence) where AttributeSequence.Element == RelativeDistinguishedName.Attribute {
+        self.attributes = .init(attributes)
+        Self._sortElements(&self.attributes)
+    }
+    
+    @inlinable
+    init(_ attributes: some Sequence<Result<RelativeDistinguishedName.Attribute, some Error>>) throws {
```
Lukasa, line 58: Do we need this to be fully generic? Or can we just accept the actual type?

dnadoba: Nope, have changed it: https://github.com/apple/swift-certificates/pull/101/commits/d2ca5723af35720530e6ef405a6a1f6743a77ce2

Lukasa (approved):

PR #106, opened 2023-06-29
Add integer initialisers on `Certificate.SerialNumber`

dnadoba:
`Certificate.SerialNumber` can now be constructed from `FixedWidthInteger` and `StaticBigInt`

### DIFF
--- a/Sources/X509/CertificateSerialNumber.swift
+++ b/Sources/X509/CertificateSerialNumber.swift
@@ -12,6 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+import SwiftASN1
+
 extension Certificate {
     /// A number that uniquely identifies a certificate issued by a specific
     /// certificate authority.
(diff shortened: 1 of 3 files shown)


Sources/X509/CertificateSerialNumber.swift
```diff
@@ -84,3 +95,34 @@ extension Certificate.SerialNumber: CustomStringConvertible {
         return self.bytes.lazy.map { String($0, radix: 16) }.joined(separator: ":")
     }
 }
+
+@available(macOS 13.3, iOS 16.4, watchOS 9.4, tvOS 16.4, *)
```
Lukasa, line 99: Note that this doesn't just want availability constraints, it needs a Swift version guard.

dnadoba: Good catch, updated

dnadoba:
Allocations have regressed, will look into that. Likely because we now always trim leading zeros but that should not allocate in our case

dnadoba:
initialising `ArraySlice` from another `ArraySlice` in a generic context will almost always copy the memory, except if the `ArraySlice` covers the whole underlying `Array`. This is because [`init(_ s: some Sequence<Element>)`](https://github.com/apple/swift/blob/1bcd73c55fa45352ec004b90b626e98c6d59d0f5/stdlib/public/core/ArraySlice.swift#L675-L679) calls [`s._copyToContiguousArray()` ](https://github.com/apple/swift/blob/1bcd73c55fa45352ec004b90b626e98c6d59d0f5/stdlib/public/core/ArraySlice.swift#L1115-L1122) which calls [`requestNativBuffer()`](https://github.com/apple/swift/blob/1bcd73c55fa45352ec004b90b626e98c6d59d0f5/stdlib/public/core/SliceBuffer.swift#L225-L231) which will only hit the fast path if the count of the slice and the underlying buffer match.


I have now added overloads which work around this limitation. I kind of which we could do something like this without adding overloads but dynamic casts are expensive if they are not specialised so we can't do that today.

Lukasa (approved):

PR #108, opened 2023-07-04
Allow `OCSPRequester` to fail `OCSPVerifierPolicy`

dnadoba:
### Motivation
A concrete `OCSPRequester` might want to decided that a request failure is fatal and therefore fail the OCSP verification.

### Modifications
Instead of throwing let the query method return a more descriptor return value that can disambiguate between terminal and non-terminal errors.

### Results
`OCSPRequester` can now result in a verification failure

### DIFF
--- a/Sources/X509/OCSP/OCSPPolicy.swift
+++ b/Sources/X509/OCSP/OCSPPolicy.swift
@@ -62,11 +62,11 @@ extension OCSPRequesterQueryResult {
     /// The OCSP query is considered unsuccessful but will **not** fail verification, neither in ``OCSPFailureMode/soft`` nor in ``OCSPFailureMode/hard`` failure mode.
     /// The certificate is then considered to meet the ``OCSPVerifierPolicy``.
     /// - Parameter reason: the reason why the OCSP query failed which may be used for diagnostics
+    /// - warning: The ``OCSPVerifierPolicy`` will assume that verification has succeeded and therefore pass OCSP verification for the given certificate.
     @inlinable
     public static func nonTerminalError(_ reason: Error) -> Self {
         .init(.nonTerminal(reason))
     }
-    
     
     /// The OCSP query is considered unsuccessful and will fail verification in both ``OCSPFailureMode/soft`` and ``OCSPFailureMode/hard`` failure mode.
     /// The certificate is then considered to not meet the ``OCSPVerifierPolicy`` and ``OCSPVerifierPolicy/chainMeetsPolicyRequirements(chain:)`` will return ``PolicyEvaluationResult/failsToMeetPolicy(reason:)`` with the given ``reason``.
--- a/Sources/X509/OCSP/OCSPPolicy.swift
+++ b/Sources/X509/OCSP/OCSPPolicy.swift
@@ -32,7 +32,40 @@ public protocol OCSPRequester: Sendable {
     ///   - request: DER-encoded request bytes
     ///   - uri: uri of the OCSP responder
     /// - Returns: DER-encoded response bytes
-    func query(request: [UInt8], uri: String) async throws -> [UInt8]
+    func query(request: [UInt8], uri: String) async -> OCSPRequesterQueryResult
+}
+
+public struct OCSPRequesterQueryResult: Sendable {
+    @usableFromInline
+    enum Storage: Sendable {
+        case success([UInt8])
+        case softFailure(Error)
+        case hardFailure(Error)
+    }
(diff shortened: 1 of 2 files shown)


Sources/X509/OCSP/OCSPPolicy.swift
```diff
@@ -31,8 +31,51 @@ public protocol OCSPRequester: Sendable {
     /// - Parameters:
     ///   - request: DER-encoded request bytes
     ///   - uri: uri of the OCSP responder
-    /// - Returns: DER-encoded response bytes
-    func query(request: [UInt8], uri: String) async throws -> [UInt8]
+    /// - Returns: DER-encoded response bytes if they request was successful or a terminal or non-terminal error.
+    func query(request: [UInt8], uri: String) async -> OCSPRequesterQueryResult
+}
+
+public struct OCSPRequesterQueryResult: Sendable {
```
FranzBusch, line 38: `Hashable`?

Sources/X509/OCSP/OCSPPolicy.swift
```diff
@@ -31,8 +31,51 @@ public protocol OCSPRequester: Sendable {
...
+}
+
+extension OCSPRequesterQueryResult {
+    /// The OCSP query is considered successful and has returned the given DER-encoded response bytes.
+    /// - Parameter bytes: DER-encoded response bytes
+    @inlinable
+    public static func response(_ bytes: [UInt8]) -> Self {
+        .init(.success(bytes))
+    }
+
```
FranzBusch, line 61:
```suggestion
```

Sources/X509/OCSP/OCSPPolicy.swift
```diff
@@ -31,8 +31,51 @@ public protocol OCSPRequester: Sendable {
...
+    init(_ storage: Storage) {
+        self.storage = storage
+    }
+}
+
+extension OCSPRequesterQueryResult {
+    /// The OCSP query is considered successful and has returned the given DER-encoded response bytes.
+    /// - Parameter bytes: DER-encoded response bytes
+    @inlinable
+    public static func response(_ bytes: [UInt8]) -> Self {
```
FranzBusch, line 58: Hm, how do you actually get the bytes out of the result with the new API? Similar question on the other APIs.

dnadoba: Its just meant to be consumed internally by `OCSPVerificationPolicy`

dnadoba:
API breakage expected:
```
15:18:17 1 breaking change detected in X509:
15:18:17   💔 API breakage: func OCSPRequester.query(request:uri:) has return type change from [Swift.UInt8] to X509.OCSPRequesterQueryResult
```

Sources/X509/OCSP/OCSPPolicy.swift
FranzBusch: Ah so nobody is going to call `query` but rather we are calling it for them. So they only need to be able to construct the return type but not handle it.

dnadoba: exactly!

FranzBusch (approved):

Sources/X509/OCSP/OCSPPolicy.swift
```diff
@@ -31,8 +31,50 @@ public protocol OCSPRequester: Sendable {
...
+        .init(.success(bytes))
+    }
+    
+    /// The OCSP query is considered unsuccessful but will **not** fail verification, neither in ``OCSPFailureMode/soft`` nor in ``OCSPFailureMode/hard`` failure mode.
+    /// The certificate is then considered to meet the ``OCSPVerifierPolicy``.
+    /// - Parameter reason: the reason why the OCSP query failed which may be used for diagnostics
+    @inlinable
+    public static func nonTerminalError(_ reason: Error) -> Self {
```
Lukasa, line 66: The doc string here should probably have a `warning` directive that calls out that if you use this the OCSP validation will be assumed to have _succeeded_.

Lukasa (changes requested):

Sources/X509/OCSP/OCSPPolicy.swift
dnadoba: Added in https://github.com/apple/swift-certificates/pull/108/commits/8c659ed5947b6de8886ee8bbb31b514aac13ea29

Lukasa (approved):

PR #112, opened 2023-07-11
Make hashing a `RDN.Attribute.Value.utf8String` not allocate

dnadoba:
### Motivation
Hashing a `RDN.Attribute.Value.utf8String` currently serialises the string to a temporary `Array`. This more than doubles allocations done during verification (allocation test not yet merged: https://github.com/apple/swift-certificates/pull/111)

### Modifications
Implement a  `String.ASN1UTF8StringView` that supports lazy iteration of the DER encoded bytes.

### Results
Allocations are reduce from ~1590 to ~660 in the allocation test introduced in #111 

`RDN.Attribute.Value.printable` will follow in a separate PR.

### DIFF
--- a/Sources/X509/RDNAttribute.swift
+++ b/Sources/X509/RDNAttribute.swift
@@ -83,9 +83,136 @@ extension RelativeDistinguishedName.Attribute.Value.Storage: Hashable {
     
     @inlinable
     func hash(into hasher: inout Hasher) {
-        // TODO: implement a ASN1UTF8StringView over a String that conforms to RandomAccessCollection and can lazily bytes that are equal to ASN1Any(ASN1UTF8String(string:))
-        // same for ASN1PrintableString. This would allow us to implement `hash(into:)` without allocating, same for default case in `==` above.
-        ASN1Any(self).hash(into: &hasher)
+        switch self {
+        case .printable:
+            // TODO: implement a ASN1PrintableStringView over a String that conforms to RandomAccessCollection and can lazily get bytes that are equal to ASN1Any(ASN1UTF8String(string:))
+            hasher.combine(ASN1Any(self))
+        case .utf8(let string):
+            hasher.combine(String.ASN1UTF8StringView(string: string))
+            
+        case .any(let asn1Any):
+            hasher.combine(asn1Any)
+        }
+    }
+}
+
+extension String {
+    @usableFromInline
+    struct ASN1UTF8StringView {
+        @usableFromInline
+        var string: String
+        
+        @inlinable
+        init(string: String) {
+            self.string = string
+        }
+    }
+}
+
+extension String.ASN1UTF8StringView: RandomAccessCollection {
+    @inlinable
+    var startIndex: Int {
+        0
+    }
+    
+    @inlinable
+    var endIndex: Int {
+        count
+    }
+    
+    @inlinable
+    var count: Int {
+        let utf8Count = self.string.utf8.count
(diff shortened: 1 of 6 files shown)


Sources/X509/RDNAttribute.swift
```diff
@@ -83,9 +83,136 @@ extension RelativeDistinguishedName.Attribute.Value.Storage: Hashable {
...
+    var endIndex: Int {
+        count
+    }
+    
+    @inlinable
+    var count: Int {
+        let utf8Count = self.string.utf8.count
+        // tag + utf8 bytes length + utf8 bytes
+        return 1 + ASN1Length(length: utf8Count).count + utf8Count
+    }
```
Lukasa, line 128: Can we compute this ahead of time? No need to recompute it each time we calculate `endIndex`.

Sources/X509/RDNAttribute.swift
```diff
@@ -83,9 +83,136 @@ extension RelativeDistinguishedName.Attribute.Value.Storage: Hashable {
...
+    
+    @inlinable
+    subscript(position: Int) -> UInt8 {
+        let length = ASN1Length(length: self.string.utf8.count)
```
Lukasa, line 132: Similarly, let's precompute this to avoid repeatedly recalculating it.

Sources/X509/RDNAttribute.swift
```diff
@@ -83,9 +83,136 @@ extension RelativeDistinguishedName.Attribute.Value.Storage: Hashable {
...
+            // We save a branch here because we can never overflow this addition.
+            let neededBits = self.length.bitWidth - self.length.leadingZeroBitCount
+            let neededBytes = (neededBits &+ 7) / 8
+            return neededBytes &+ 1
+        }
+    }
+    
+    @inlinable
+    subscript(position: Int) -> UInt8 {
+        precondition(position >= 0 && position < self.length)
```
Lukasa, line 201: Presumably here `position` was supposed to be less than `count`, not `length`?

Sources/X509/RDNAttribute.swift
```diff
@@ -83,9 +83,136 @@ extension RelativeDistinguishedName.Attribute.Value.Storage: Hashable {
...
+            return 1
+        } else {
+            // We need to work out how many bytes we need. There are many fancy bit-twiddling
+            // ways of doing this, but honestly we don't do this enough to need them, so we'll
+            // do it the easy way. This math is done on UInt because it makes the shift semantics clean.
+            // We save a branch here because we can never overflow this addition.
+            let neededBits = self.length.bitWidth - self.length.leadingZeroBitCount
+            let neededBytes = (neededBits &+ 7) / 8
+            return neededBytes &+ 1
+        }
```
Lukasa, line 196: We can precompute this too.

Lukasa (changes requested):
A few notes, but also: when is this called on the verification path?

dnadoba:
Good catches! Applied them in https://github.com/apple/swift-certificates/pull/112/commits/dd1c74edbe48f4b0e54c386e2c4aabd54555fa0f
This is called for all root and intermediate certificates at least once because their subject name is used in `CertificateStore` as keys in the `Dictionary`. During verification the certificates issuer is looked up in the `CertificateStore`s which requires hashing of this again:
https://github.com/apple/swift-certificates/blob/c13badf75a49f052102188d4ca98cc15dc853f6f/Sources/X509/Verifier/Verifier.swift#L61

Lukasa (approved):

PR #113, opened 2023-07-12
Add optional diagnostic information to `Verifier`

dnadoba:
### Motivation
It can be hard why a given set of roots and intermediate certificates don't produce a valid certificate chain. Even if it found a chain it might still not find the shorts or the specific chain that one would expect. We should provide a way to expose the information that is already available inside the verifier to the user.

### Modifications
- Change `diagnosticCallback: ((String) -> Void)? = nil` from a `String` to a new type `VerificationDiagnostic` that allows to delay computation of the description to an arbitrary point in time. It is also `Sendable` which allows to put the computation in on a different thread.
- Produce detailed diagnostics information during verification, but only if the user provided the `diagnosticCallback`. 

### Result
Diagnostics can be forwarded to the user. They are zero cost if `diagnosticCallback` is `nil` and lightweight if `VerificationDiagnostic.description` is not called.


Sources/X509/Verifier/VerificationDiagnostic.swift
```diff
@@ -0,0 +1,237 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the SwiftCertificates open source project
+//
+// Copyright (c) 2022 Apple Inc. and the SwiftCertificates project authors
```
glbrntt, line 5:
```suggestion
// Copyright (c) 2023 Apple Inc. and the SwiftCertificates project authors
```

Sources/X509/Verifier/VerificationDiagnostic.swift
```diff
@@ -0,0 +1,237 @@
...
+        case intermediateHashUnhandledCriticalExtension(IssuerHasUnhandledCriticalExtension)
+        case issuerHasNotSignedCertificate(IssuerHasNotSignedCertificate)
+    }
+    var storage: Storage
+}
+
+extension VerificationDiagnostic {
+    static func leafCertificateHasUnhandledCriticalExtension(
+        _ leafCertificate: Certificate,
+        handledCriticalExtensions: [ASN1ObjectIdentifier]
```
glbrntt, line 51: As a diagnostic for unhandled extensions I would expect to see whatever isn't handled but this is `handled`... is that a typo?

Sources/X509/Verifier/VerificationDiagnostic.swift
```diff
@@ -0,0 +1,237 @@
...
+    static func chainFailsToMeetPolicy(
+        _ chain: UnverifiedCertificateChain,
+        reason failsToMeetPolicyReason: String
+    ) -> Self {
+        self.init(storage: .chainFailsToMeetPolicy(
+            chain,
+            reason: failsToMeetPolicyReason
+        ))
+    }
+    static func issuerHashUnhandledCriticalExtension(
```
glbrntt, line 76: hash or has?

Sources/X509/Verifier/VerificationDiagnostic.swift
```diff
@@ -0,0 +1,237 @@
...
+    static func chainFailsToMeetPolicy(
+        _ chain: UnverifiedCertificateChain,
+        reason failsToMeetPolicyReason: String
+    ) -> Self {
+        .chainFailsToMeetPolicy(.init(
+            chain: chain,
+            failsToMeetPolicyReason: failsToMeetPolicyReason
+        ))
+    }
+    static func issuerHashUnhandledCriticalExtension(
```
glbrntt, line 126: same here: has or hash?

Sources/X509/Verifier/VerificationDiagnostic.swift
```diff
@@ -0,0 +1,237 @@
...
+    ) -> Self {
+        .issuerHasNotSignedCertificate(.init(
+            issuer: issuer,
+            partialChain: partialChain
+        ))
+    }
+}
+
+extension VerificationDiagnostic: CustomStringConvertible {
+    /// Produces a human readable description of this ``VerificationDiagnostic`` that is potential expensive to compute.
```
glbrntt, line 149:
```suggestion
    /// Produces a human readable description of this ``VerificationDiagnostic`` that is potentially expensive to compute.
```

Sources/X509/Verifier/VerificationDiagnostic.swift
```diff
@@ -0,0 +1,237 @@
...
+        case .intermediateHashUnhandledCriticalExtension(let diagnostic): return diagnostic.description
+        case .issuerHasNotSignedCertificate(let diagnostic): return diagnostic.description
+        }
+    }
+}
+
+extension VerificationDiagnostic.LeafCertificateHasUnhandledCriticalExtensions: CustomStringConvertible {
+    var description: String {
+        """
+        The leaf certificate has critical extensions that the policy does not understand and therefore can't enforce.
```
glbrntt, line 170: Given this is multiline output, it might make sense to add new lines in appropriate places to make it more readable:

```swift
"""
The leaf certificate ... enforce.

Unhandled extensions:
...

Leaf certificate:
...
"""
```

Further to that I don't know what the format of the unhandled extensions is but might make sense to bullet point each.

Sources/X509/Verifier/VerificationDiagnostic.swift
```diff
@@ -0,0 +1,237 @@
...
+        Leaf certificate:
+        \(leafCertificate)
+        """
+    }
+}
+
+extension VerificationDiagnostic.LeafCertificateIsInTheRootStoreButDoesNotMeetPolicy: CustomStringConvertible {
+    var description: String {
+        """
+        Leaf certificate is in the root store of the verifier but it does by itself not meet the policy.
```
glbrntt, line 184:
```suggestion
        Leaf certificate is in the root store of the verifier but it does not meet the policy.
```

Sources/X509/Verifier/VerificationDiagnostic.swift
```diff
@@ -0,0 +1,237 @@
...
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftCertificates project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+import SwiftASN1
+
+public struct VerificationDiagnostic: Sendable {
```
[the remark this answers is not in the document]
dnadoba: Yes, I don't think there is currently a reason why they would want to. We do not guarantee what diagnostics we produce and don't want to. `Verifier` wants to find a valid chain as fast as possible and doesn't check all possible paths as it would be rather expensive (and in theory impossible because it can contains loops). This means that the diagnostics can and will change if we improve the graph traversal algorithm or other parts of the verifier. `VerificationDiagnostic` is also not `Equatable` for the same reason.

Sources/X509/Verifier/VerificationDiagnostic.swift
dnadoba: Good catch! it should be "has"

Sources/X509/Verifier/VerificationDiagnostic.swift
dnadoba: Yep, lets do that.

Sources/X509/Verifier/VerificationDiagnostic.swift
dnadoba: Not a typo but on purpose. We delay the computation of the unhandled extensions until `description` is accessed to make constructing a `LeafCertificateHasUnhandledCriticalExtension` just a couple of retains. `handledCriticalExtensions` and `leaf` are already constructed when we produce the diagnostics. Unhandled extensions are computed in the `Certificate.Extensions.unhandledCriticalExtensions(for:)` at the very bottom of this file.

Sources/X509/Verifier/VerificationDiagnostic.swift
dnadoba: I kinda want to highlight that the leaf certificate may only not meet the policy if the whole chain is just this single certificate but may pass the policy if we try a different chain e.g. through intermediates. We don't know that yet though, just that it is in the root store and the policy has failed on that single certificate chain. We will continue trying to find new paths through intermediates to a root which can then still pass the policy.

glbrntt (approved):
Useful errors and diagnostics are really valuable; this is great!

Sources/X509/Verifier/Verifier.swift
```diff
@@ -113,6 +124,7 @@ public struct Verifier<Policy: VerifierPolicy> {
 
         // We check the signature here: if the signature isn't valid, don't try to apply policy.
         guard nextCertificate.publicKey.isValidSignature(partialChain.currentTip.signature, for: partialChain.currentTip) else {
+            diagnosticCallback?(.issuerHasNotSignedCertificate(nextCertificate, chain: partialChain))
             return true
         }
```
Lukasa, line 130: We're missing two diagnostics in this function for two other returns.

Lukasa (approved):

PR #118, opened 2023-08-09
Reject `v1` `Certificate`s with extensions

dnadoba:
v1 certificates do not support extensions and any v1 certificate we encounter with extensions can’t be used to verify a chain.


Lukasa (approved):
Thanks! We may want in the future to be able to distinguish the difference between not having extensions, and having empty extensions, but for now I think this is totally sufficient.

PR #122, opened 2023-08-16
Allow `RelativeDistinguishedNameConvertible` to fail gracefully

dnadoba:
Port the changes over from https://github.com/apple/swift-certificates/pull/80 to `DistinguishedName`
This slightly breaks API as we need to change the return type from `DistinguishedName` to `Result<DistinguishedName, any Error>` but is necessary to gracefully handle errors thrown from the result builder.


Lukasa (approved):
Nice one, LGTM.

PR #125, opened 2023-08-18
Adopt `package-benchmark`

dnadoba:


benchmarks/Sources/VerifierBenchmark.swift
```diff
@@ -0,0 +1,590 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import X509
+import Foundation
+import Crypto
+import SwiftASN1
+
+public func verifier() async {
```
FranzBusch, line 20: I would love if we separate each benchmark into a separate file and move the supporting code into separate files as well. I don't wanna end up with a massive file again like the current NIO benchmarks have.

benchmarks/Sources/ParseWebPKI.swift
```diff
@@ -0,0 +1,158 @@
...
+// See CONTRIBUTORS.txt for the list of SwiftCertificates project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import X509
+import SwiftASN1
+
+import Foundation
```
FranzBusch, line 18:
```suggestion
import Foundation

```

benchmarks/Benchmarks/CertificatesBenchmarks/Benchmarks.swift
```diff
@@ -0,0 +1,35 @@
...
+        runParseWebPKIRoots()
+    }
+    
+    Benchmark("TinyArray non-allocating functions") { benchmark in
+        tinyArrayNonAllocationFunctions()
+    }
+    
+    Benchmark("TinyArray.append(_:)") { benchmark in
+        tinyArrayAppend()
+    }
```
FranzBusch, line 34: This uses the default metrics now, we should either use `all` or select the metrics we care about.

benchmarks/Benchmarks/CertificatesBenchmarks/Benchmarks.swift
```diff
@@ -0,0 +1,35 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Benchmark
+import Sources
+
+let benchmarks = {
+    Benchmark("Verifier", configuration: .init(warmupIterations: 1)) { benchmark in
```
FranzBusch, line 19: Why do we warm up here but not in the others?

FranzBusch (commented):
Overall love it! We should also commit the base lines and let all of this run on the CI. We probably don't want to have the time based metrics run on the CI.

benchmarks/Sources/VerifierBenchmark.swift
dnadoba: It is a single benchmark. Not sure if we want separate benchmarks per validation use case. Maybe successful and unsuccessful as separate benchmarks but that's it.

benchmarks/Benchmarks/CertificatesBenchmarks/Benchmarks.swift
dnadoba: We do access a bunch of `static let`s that are initialised on first access. "Parse WebPKI Roots" should get it too but the others have not shown any benefit. But it also doesn't hurt AFAIK so I'm open to add it everywhere.

benchmarks/Sources/ParseWebPKI.swift
dnadoba: I will add this folder to our formatting as well

benchmarks/Sources/blackHole.swift
```diff
@@ -0,0 +1,31 @@
...
+///         metrics: [.throughput, .wallClock],
+///         scalingFactor: .mega)
+/// ) { benchmark in
+///     for _ in benchmark.scaledIterations {
+///         blackHole(Date())
+///     }
+/// }
+/// ```
+@inline(never)
+public func blackHole(_: some Any) {}
```
hassila, line 31: This blackHole might be a bit problematic? It'd be compiled together in the same module as the code under measurement, so it might not work as expected, or am I missing something? Shouldn't it go to a separate target/module ?

benchmarks/Benchmarks/CertificatesBenchmarks/Benchmarks.swift
FranzBusch: Might be worth adding a comment for that.

benchmarks/Benchmarks/CertificatesBenchmarks/Benchmarks.swift
hassila: I'd suggest you set up 

```swift
Benchmark.defaultConfiguration = .init(...)
```

at the top of the suite with whatever default config you want to use for all benchmarks (if the same), then you can easily add metrics for the whole suite without resorting to ".all" which may be overkill for most uses.

If you have subsets that needs another config just do:
```swift
    let someCustomConfiguration: Benchmark.Configuration = .init(...)

    Benchmark("xxx 1", configuration: someCustomConfiguration) { _ in
    }

    Benchmark("xxx 2", configuration: someCustomConfiguration) { _ in
    }
```

Also convenient for setting up thresholds.

benchmarks/Benchmarks/CertificatesBenchmarks/Benchmarks.swift
```diff
@@ -0,0 +1,35 @@
...
+        await verifier()
+    }
+    
+    let runParseWebPKIRoots = parseWebPKIRoots()
+    Benchmark("Parse WebPKI Roots") { benchmark in
+        runParseWebPKIRoots()
+    }
+    
+    Benchmark("TinyArray non-allocating functions") { benchmark in
```
hassila, line 28: For the kind of very small (short run time) tests, I would suggest you use the support for `scaledIterations` with an inner loop to decrease the benchmark measurement overhead and get higher quality measurements, e.g.:

```swift
    Benchmark.defaultConfiguration.scalingFactor = .kilo
...
    Benchmark("TinyArray non-allocating functions") { benchmark in
        for _ in benchmark.scaledIterations {
            tinyArrayNonAllocationFunctions()
        }
    }
```
You can get the time measurements (and malloc stats etc) scaled to per-iteration using i.e.
```
swift package benchmark --scale
```

hassila:
> Overall love it! We should also commit the base lines and let all of this run on the CI. We probably don't want to have the time based metrics run on the CI.

Yeah, time based metrics mostly only works if you have dedicated CI runners for them (that's how we do it), the results are way to flakey in general at least on GitHubs runners. But you might still want to run the metrics on the CI and just set thresholds that allow for any time deviation to be ok, then you have them readily accessible if/when running and checking things manually.

benchmarks/Sources/blackHole.swift
hassila: One thing to look for is if you get a runtime that is extremely short, on my machine that benchmark executes on a wall clock time of ~41ns - exactly the same as I'll get if I simply comment out the call to tinyArrayNonAllocationFunctions() - even if I put in an inner loop there, it still takes ~41ns (which seems to be the overhead of a function call?).

I'd probably suggest a small restructuring something like this:
```swift
import _CertificateInternals

public func tinyArrayNonAllocationFunctions(_ iterations: Range<Int>) -> Int {
    var counts = 0
    for _ in iterations {
        counts += _TinyArray(CollectionOfOne(1)).count

        var array = _TinyArray<Int>()
        array.append(contentsOf: CollectionOfOne(1))
        counts += array.count
    }
    return counts
}

with a call site from the benchmark code:

    Benchmark("TinyArray non-allocating functions") { benchmark in
        blackHole(tinyArrayNonAllocationFunctions(benchmark.scaledIterations)) // this uses the benchmarks blackHole
    }
```

hassila: See also https://github.com/ordo-one/package-benchmark/issues/178

benchmarks/Thresholds/5.7/CertificatesBenchmarks.Parse_WebPKI_Roots.p90.json
```diff
@@ -0,0 +1,30 @@
+[
+  {
```
FranzBusch, line 2: I find this format a bit unclear tbh. I expected something along those lines

```
{
  "retain" {
    "count": XXX
  },
  "allocations": {
    "total": XXX,
    "large": XXX,
    "small": XXX,
  }
}
```

@hassila Is this something we could get to work?

benchmarks/Benchmarks/CertificatesBenchmarks/Benchmarks.swift
```diff
@@ -0,0 +1,80 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Benchmark
+import Sources
+import Foundation
+
+func makeConfigurationFor(_ name: String) -> Benchmark.Configuration {
```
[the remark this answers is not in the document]
hassila: Give latest https://github.com/ordo-one/package-benchmark/pull/180 branch a whirl:

```bash
swift package benchmark baseline check --check-absolute-path Thresholds
```

image

The sample code for P90 tests is now simply the clean:

```swift
import Benchmark
import Foundation

let benchmarks = {
    Benchmark.defaultConfiguration = .init(metrics: [.mallocCountTotal, .syscalls],
                                           warmupIterations: 1,
                                           scalingFactor: .kilo,
                                           maxDuration: .seconds(2),
                                           maxIterations: .kilo(100))

    Benchmark("P90Date") { benchmark in
        for _ in benchmark.scaledIterations {
            blackHole(Foundation.Date())
        }
    }

    Benchmark("P90Malloc") { benchmark in
        for _ in benchmark.scaledIterations {
            var array: [Int] = []
            array.append(contentsOf: 0 ... 1_000)
            blackHole(array)
        }
    }
}
```

benchmarks/Thresholds/5.7/CertificatesBenchmarks.Parse_WebPKI_Roots.p90.json
hassila: The format is basically an implementation detail, it's machine generated and machine read?

FranzBusch: I think that's the difference here for us a bit. We want to understand from the file what the different thresholds are so that if somebody just updates the thresholds we understand if we regressed allocations and by how much.

hassila: https://github.com/ordo-one/package-benchmark/pull/180/commits/0b2a4d853427458cf60739feb04fee30f337eb3c

hassila: Should be more readable, it's basically:

```bash
hassila@ice ~/g/package-benchmark (threshold-export)> more Thresholds/P90AbsoluteThresholdsBenchmark.P90Date.p90.json
{
  "syscalls" : 1,
  "mallocCountTotal" : 0
}
hassila@ice ~/g/package-benchmark (threshold-export)> 
```

FranzBusch: Oh yeah that looks lovely! Thanks for the quick turn around on those. Very much appreciated and I think we are now at stage where we can actually finalise the PR. Right @dnadoba ?

hassila:
Just one heads-up for the arc numbers - in case you do other tests that use task groups: https://github.com/apple/swift/issues/64636 (still relevant to check for most tests, just FYI)

FranzBusch:
> Just one heads-up for the arc numbers - in case you do other tests that use task groups: https://github.com/apple/swift/issues/64636 (still relevant to check for most tests, just FYI)

@ktoso the issue might be more relevant now when we adopt the benchmark plugin

benchmarks/Benchmarks/CertificatesBenchmarks/Benchmarks.swift
```diff
@@ -0,0 +1,49 @@
...
+//===----------------------------------------------------------------------===//
+
+import Benchmark
+import Sources
+import Foundation
+
+let benchmarks = {
+    Benchmark.defaultConfiguration = .init(
+        metrics: [.mallocCountTotal, .syscalls] + .arc,
+        warmupIterations: 1
```
simonjbeaumont, line 22: FYI this is the default already.

benchmarks/.gitignore
```diff
@@ -0,0 +1,8 @@
+.DS_Store
```
FranzBusch, line 1: I think this file can be removed

benchmarks/Package.swift
```diff
@@ -0,0 +1,59 @@
...
+            ],
+            path: "Sources"
+        ),
+        .testTarget(
+            name: "Tests",
+            dependencies: [
+                "Sources",
+            ],
+            path: "Tests"
+        )
```
FranzBusch, line 57: I find the naming and paths here a bit weird. Can't we come up with something in the default path format?

benchmarks/Tests/Tests.swift
```diff
@@ -0,0 +1,48 @@
...
+//===----------------------------------------------------------------------===//
+
+import Sources
+import XCTest
+
+final class TestRunner: XCTestCase {
+    override func setUpWithError() throws {
+        #if DEBUG
+        throw XCTSkip("performance tests only run in release mode")
+        #endif
```
FranzBusch, line 22: Maybe we even want to `fatalError` here?

scripts/run-swift-format.sh
```diff
@@ -39,14 +39,15 @@ swiftformat_bin=${swiftformat_bin:-$(command -v swift-format)} || fatal "❌ swi
 "${swiftformat_bin}" lint \
   --parallel --recursive --strict \
   "${repo_root}/Sources" "${repo_root}/Tests" \
+    "${repo_root}/benchmarks/Benchmarks" "${repo_root}/benchmarks/Sources" "${repo_root}/benchmarks/Tests" \
```
FranzBusch, line 42: Can't we just format `${repo_root}/benchmarks/`?

.gitignore
```diff
@@ -11,3 +11,4 @@ Package.resolved
 .swiftpm
 /out
 
+/benchmarks/.benchmarkBaselines
```
FranzBusch, line 14: Do we still need this ignore?

benchmarks/.gitignore
dnadoba: No, we need it as `.build` is otherwise included. I don't actually mind this additional .gitignore

benchmarks/Package.swift
dnadoba: Changed

benchmarks/Tests/Tests.swift
dnadoba: yeah, lets do this.

.gitignore
FranzBusch: Ah gotcha. I confused baselines with thresholds.

scripts/run-swift-format.sh
FranzBusch: I think that should work. Will also format the `Package.swift` then

benchmarks/.gitignore
FranzBusch: I think we should add that to the root `.gitingore` instead of having nested ones. WDYT?

dnadoba: I'm not a fan of that. If we ever going to remove the benchmarks folder we would need to remember to edit the parent `.gitignore`. I quite like that git can have nested gitignores and I don't really see a downside to it. Why do you think this should live in the parent `.gitignore`?

scripts/run-swift-format.sh
dnadoba: Just tested and it doesn't work because it will also try to format hidden folders and therefore our checked out dependencies e.g.:
```
~/Repositories/swift-format/.build/release/swift-format format --parallel --recursive --in-place ./                                                                     
/Users/davidnadoba/Repositories/swift-certificates/benchmarks/.build/5.8/checkouts/package-benchmark/Sources/Benchmark/Benchmark.swift:18:46: error: expected declaration after attribute in conditional compilation clause
```

FranzBusch: Yeah I think what I would recommend is inverting this (don't know if `swift-format` supports that) e.g. passing excludes instead of includes

benchmarks/.gitignore
FranzBusch: I don't have a strong opinions. I personally find nested `.gitignores` a bit weird.

scripts/run-swift-format.sh
dnadoba: It doesn't support that

dnadoba:
Anything else? @FranzBusch @Lukasa

hassila:
Just another FYI, if you are interested in checking bytes read/written to disk, we just added support for macOS now too (Linux was already supported). Maybe not important for this repo, but perhaps for others.

https://github.com/ordo-one/package-benchmark/releases/tag/1.9.1

benchmarks/Benchmarks/BenchmarksRunner/Benchmarks.swift
```diff
@@ -0,0 +1,49 @@
...
+            await verifier()
+        }
+    }
+
+    let runParseWebPKIRoots = parseWebPKIRoots()
+    Benchmark("Parse WebPKI Roots") { benchmark in
+        for _ in benchmark.scaledIterations {
+            runParseWebPKIRoots()
+        }
+    }
```
simonjbeaumont, line 36: I think the pattern for doing setup _inside_ the closure might be cleaner because you don't need to worry about conflicts in the outer scope. You do this by deferring the start of the measurement like this, IIUC:

```suggestion
    Benchmark("Parse WebPKI Roots") { benchmark in
        let runParseWebPKIRoots = parseWebPKIRoots()
        benchmark.startMeasurement()
        for _ in benchmark.scaledIterations {
            runParseWebPKIRoots()
        }
    }
```

simonjbeaumont (commented):
Really happy to see the use of package-benchmarks. IIUC you are using three targets:

1. A target with some public functions that have some hot loops for benchmarking;
2. A benchmark target that calls the functions in (1); and
3. A test target that calls the functions in (1).

If this is just to enable a workflow for profiling in Xcode I don't think this is necessary and I have found that all can be accomplished by rolling everything into (2), which also allows you to make use of `Benchmark` utilities for scaling the loops etc.

There is an open issue about using jemalloc in Xcode, but you can disable it if you need https://github.com/ordo-one/package-benchmark/issues/163.

Or, you can run `xctrace` on the command line and then open in Xcode:

```
❯ xctrace record --output foo.trace --append-run --template "Time Profiler" --launch -- $(swift build -c release --show-bin-path)/MyBenchmarks
❯ open foo.trace
```

hassila:
> There is an open issue about using jemalloc in Xcode, but you can disable it if you need [ordo-one/package-benchmark#163](https://github.com/ordo-one/package-benchmark/issues/163).

In case anyone is interested, I also filed these two related issues:

FB12061292
and
FB12076783

benchmarks/Benchmarks/BenchmarksRunner/Benchmarks.swift
hassila: I'd suggest instead that you consider:

```suggestion
    Benchmark("Parse WebPKI Roots") { benchmark in
        for _ in benchmark.scaledIterations {
            runParseWebPKIRoots()
        }
    } setup: {
       parseWebPKIRoots()
    } 
```

Otherwise it'd be run for each iteration of the benchmark.

`scaledIterations` is the inner loop of the benchmark, but each benchmark function can also be called multiple times depending on run time and number of desired iterations, so using custom setup/teardown is intended to support such use cases. More details in [the documentation](https://swiftpackageindex.com/ordo-one/package-benchmark/1.9.1/documentation/benchmark/writingbenchmarks)

hassila:
> 1. A target with some public functions that have some hot loops for benchmarking;
> 2. A benchmark target that calls the functions in (1); and
> 3. A test target that calls the functions in (1).
> 
> If this is just to enable a workflow for profiling in Xcode I don't think this is necessary and I have found that all can be accomplished by rolling everything into (2), which also allows you to make use of `Benchmark` utilities for scaling the loops etc.


Also, to expand on this a bit: 

It's possible to debug in Xcode the individual benchmark (if you haven't seen it) by just opening the Package.swift and selecting the individual benchmark target:

image

Basically, there is a simplified built-in benchmark runner that allows for easy debugging this way.

The  problem with profiling this target is due to Xcode trying to build for x86 and arm when profiling and you'll get a linker error `ld: warning: ignoring file '/opt/homebrew/Cellar/jemalloc/5.3.0/lib/libjemalloc.2.dylib': found architecture 'arm64', required architecture 'x86_64'` It's basically due to the bug outlined here: https://forums.swift.org/t/restrict-macos-builds-to-x86-64-with-spm/42239/15 which hopefully will be addressed sometime (right now, any release build takes 2x the required time due to Xcode building for all platforms regardless of the settings in the project)

The workaround is as @simonjbeaumont mentions to disable jemalloc when profiling the target in the linked case (but for debugging it works as-is).

benchmarks/Benchmarks/BenchmarksRunner/Benchmarks.swift
dnadoba: The `setup` closure would need to allow me to return a value, in this case the `runParseWebPKIRoots` closure,  that is then passed into the benchmark function so I can actually run the benchmark. Otherwise I need to use some global storage.

hassila: Ok, maybe it would make sense to have some state holder for the benchmark that can be accessed from setup/teardown to support that use case to avoid having to rerun it in the preamble.

ktoso:
> > Just one heads-up for the arc numbers - in case you do other tests that use task groups: [apple/swift#64636](https://github.com/apple/swift/issues/64636) (still relevant to check for most tests, just FYI)
> 
> @ktoso the issue might be more relevant now when we adopt the benchmark plugin

I looked into this and it's not a bug afaics; Please see https://github.com/apple/swift/issues/64636

benchmarks/Benchmarks/BenchmarksRunner/Benchmarks.swift
hassila: The setup closure can now return its state: https://github.com/ordo-one/package-benchmark/releases/tag/1.10.0

hassila:
Also see https://github.com/ordo-one/package-benchmark/releases/tag/1.10.0 for suggestion on easier workflow for profiling and the addition of optional checks in CI for improvements on thresholds.

FranzBusch (approved):
LGTM!

benchmarks/Benchmarks/CertificatesBenchmark/Benchmarks.swift
```diff
@@ -17,10 +17,21 @@ import Foundation
...
 let benchmarks = {
     Benchmark.defaultConfiguration = .init(
-        metrics: [.mallocCountTotal, .syscalls] + .arc
+        metrics: [
+            .mallocCountTotal,
+            .syscalls,
+            .readSyscalls,
+            .writeSyscalls,
+            .memoryLeaked,
+            .retainCount,
```
hassila, line 26: one of these should be `.releaseCount` surely

hassila: (also need to rerun the threshold generation, as .releaseCount was gone now)

dnadoba: fixed in https://github.com/apple/swift-certificates/pull/125/commits/05943cb0badcfaed46b447696263d6b1e39f3caa

FranzBusch (commented):
@dnadoba I have one more request. Can we add something to our readme here that links to the benchmark docs for how to get started with the benchmarking? Just a two liner or something like this so people find how to run and debug benchmarks.

README.md
```diff
@@ -48,3 +48,24 @@ dependencies: [
 ```
 
 For detailed usage and API documentation, check [the documentation](https://swiftpackageindex.com/apple/swift-certificates/main/documentation/x509).
+
+## Benchmarks
+
+Benchmarks for `swift-certificates` are in a separate Swift Package in the `benchmarks` subfolder of this repository.
```
FranzBusch, line 54: Oh I have one more nit: Can we name the folder `Benchmarks` to be in-line with `Sources` and `Tests`?

dnadoba:
@FranzBusch added a section to the README

README.md
```diff
@@ -48,3 +48,24 @@ dependencies: [
 ```
 
 For detailed usage and API documentation, check [the documentation](https://swiftpackageindex.com/apple/swift-certificates/main/documentation/x509).
+
+## Benchmarks
+
+Benchmarks for `swift-certificates` are in a separate Swift Package in the `benchmarks` subfolder of this repository. 
+They use the [`package-benchmark`](https://github.com/ordo-one/package-benchmark) plugin.
+Benchmarks depends on the [`jemalloc`](https://jemalloc.net) memory allocation library, which is used by `package-benchmark` to capture memory allocation statistics.
+An installation guide can be found in the [Getting Started article](https://swiftpackageindex.com/ordo-one/package-benchmark/1.11.1/documentation/benchmark/gettingstarted#Installing-Prerequisites-and-Platform-Support) of `package-benchmark`.
```
FranzBusch, line 57: This part is very prone to get outdated. Can we just link to the index page of the documentation here.

README.md
```diff
@@ -48,3 +48,24 @@ dependencies: [
...
+Afterwards you can run the benchmarks from CLI by going to the `benchmarks` subfolder (e.g. `cd benchmarks`) and invoking:
+```
+swift package benchmark
+```
+
+Profiling benchmarks in Xcode (or building the benchmarks in release mode) with `jemalloc` is currently not supported and requires disabling `jemalloc`. 
+Make sure Xcode is closed and then open it from the CLI with the `BENCHMARK_DISABLE_JEMALLOC=true` environment variable set e.g.:
+```
+BENCHMARK_DISABLE_JEMALLOC=true xed .
+```
```
FranzBusch: Do they have this documented as well? I would prefer if we just keep as much out of our docs as possible.

README.md
dnadoba: its on purpose not uppercase as it is a separate package and folders of packages are lowercase.

README.md
dnadoba: I have removed the version number and it will now always pointer to the latest version. I really want to keep the direct link to the Getting Started article. I don't think it is very likely they will rename the Getting Started article. The section name  "Installing-Prerequisites-and-Platform-Support" may change but that's fine, the link will still work and point to the top of the getting started page.

README.md
dnadoba: No they haven't. I'm fine to remove it if `package-benchmark` adds it but we can also remove it again in a subsequence PR to unblock more benchmarks that I have in the pipeline. cc @hassila

FranzBusch (approved):
Nice! Thanks for pushing on this.

docker/docker-compose.2204.57.yaml
```diff
@@ -12,7 +12,8 @@ services:
   test:
     image: swift-certificates:22.04-5.7
     environment:
-      - MAX_ALLOCS_ALLOWED_parse_webpki_roots=200050
+      - SWIFT_VERSION=5.7
+      - MAX_ALLOCS_ALLOWED_parse_webpki_roots=422050
       - MAX_ALLOCS_ALLOWED_tiny_array_cow_append_contents_of=9050
```
Lukasa, line 17: Any reason not to drop the `MAX_ALLOCS_ALLOWED` variables?

README.md
hassila: We can definitely add the comment about how to run in Xcode for profiling etc, np.

docker/docker-compose.2204.57.yaml
dnadoba: No, just an oversight. removed them now.

Lukasa (approved):

PR #116, opened 2023-08-08
Add custom `String` representation to `Certificate` and related types

dnadoba:
Examples:
```swift
"""
Certificate(\
version: X509v3, \
serialNumber: 1:2:3:4:5:6:7:8:9:a, \
issuer: "CN=Swift Certificate Test CA 1,O=Apple,C=US", \
subject: "CN=Swift Certificate Test CA 1,O=Apple,C=US", \
notValidBefore: 2022-08-08 14:26:14 +0000, \
notValidAfter: 2033-08-05 14:26:14 +0000, \
publicKey: P384, \
signature: ECDSA, \
extensions: [\
BasicConstraints(CA=TRUE), \
KeyUsage(keyCertSign), \
SubjectKeyIdentifier(56:62:b1:ff:3e:a6:e2:44:d8:97:4c:c7:a:fa:68:47:92:a3:23:fb)\
]\
)
"""
```
```swift
"""
Certificate(\
version: X509v3, \
serialNumber: 1:2:3:4:5:6:7:8:9:a:b, \
issuer: "CN=Swift Certificate Test CA 1,O=Apple,C=US", \
subject: "CN=Swift Certificate Test Intermediate CA 1,O=Apple,C=US", \
notValidBefore: 2022-08-08 14:26:14 +0000, \
notValidAfter: 2028-08-06 14:26:14 +0000, \
publicKey: P256, \
signature: ECDSA, \
extensions: [\
BasicConstraints(CA=TRUE, maxPathLength=1), \
KeyUsage(keyCertSign), \
AuthorityKeyIdentifier(keyID: 56:62:b1:ff:3e:a6:e2:44:d8:97:4c:c7:a:fa:68:47:92:a3:23:fb), \
SubjectKeyIdentifier(56:62:b1:ff:3e:a6:e2:44:d8:97:4c:c7:a:fa:68:47:92:a3:23:fb), \
NameConstraints(\
permittedSubtrees: [DNSName("apple.com."), IPAddress([127, 0, 0, 0, 0, 0, 0, 255]), RFC822Name("foo@exmaple.com."), RFC822Name("bar@example.com."), URI(".example.com")], \
excludedSubtrees: [DNSName("www.apple.com."), IPAddress([127, 0, 0, 1]), RFC822Name("bar@example.com."), URI(".foo.example.com")]\
)\
])
"""
```
```swift
"""
Certificate(\
version: X509v3, \
serialNumber: 1:2:3:4:5:6:7:8:9:a:b, \
[...]

PR #126, opened 2023-08-21
Add `PolicyFailureReason` to support lazy failure reason string interpolation

dnadoba:
### Motivation 
Allocations have increased significantly during verification as noted in [this PR](https://github.com/apple/swift-certificates/pull/116#issuecomment-1669868898). The reason for that was that policy failures often use string interpolation to build dynamic strings which can allocate quite a bit, especially if values with complicated string representations are used, such as `Certificate`s.

### Changes
Instead of a `String`, take an `@autoclosure () -> String` that delays the actual string interpolation until the policy failure is actually converted to a `String`.

### Result
Allocations are reduced again if the policy failure is not printed.


Lukasa (approved):
Nice one, LGTM.

PR #127, opened 2023-08-22
Fix preference of root certificates

dnadoba:
### Motivation
We prefer issuers that have a subject key identifier that matches the authority key identifier of the certificate for which were are searching for an issuer. However, currently we have flipped this sort order.

### Modifications
- Flip preference order
- update test
- add test which would previously fail

### Result
We actually prefer certificates in the order described above. 
Note that this is just a performance improvement.

### DIFF
--- a/Sources/X509/Verifier/Verifier.swift
+++ b/Sources/X509/Verifier/Verifier.swift
@@ -120,6 +120,9 @@ public struct Verifier<Policy: VerifierPolicy> {
                     )
                 )
 
+                // we need to reverse the order of the already sorted intermediates because
+                // we will push them on to the `partialChains` stack which in turn will
+                // consume them in the reverse order the have been pushed onto the stack
                 for parent in intermediateParents.reversed() {
                     if self.shouldSkipAddingCertificate(
                         partialChain: nextPartialCandidate,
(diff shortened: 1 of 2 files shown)


FranzBusch (approved):

Sources/X509/Verifier/Verifier.swift
```diff
@@ -120,7 +120,7 @@ public struct Verifier<Policy: VerifierPolicy> {
                     )
                 )
 
-                for parent in intermediateParents {
+                for parent in intermediateParents.reversed() {
```
Lukasa, line 123: Can we document why we reverse this? (Answer: we push these onto a stack so the order in which we iterate them here is the _reverse_ order in which we explore them).

Sources/X509/Verifier/Verifier.swift
```diff
@@ -120,7 +120,10 @@ public struct Verifier<Policy: VerifierPolicy> {
                     )
                 )
 
-                for parent in intermediateParents {
+                // we need to reverse the order of the already sorted intermediates because
+                // we will push them on to the `partialChains` stack which in turn will
+                // consume them in the reverse order the have been pushed onto the stack
```
Lukasa, line 125:
```suggestion
                // consume them in the reverse order that they have been pushed onto the stack
```

Tests/X509Tests/VerifierTests.swift
```diff
@@ -556,6 +575,42 @@ final class VerifierTests: XCTestCase {
         )
     }
 
+    func testRootsWithSPIArePreferred() async throws {
```
Lukasa, line 578:
```suggestion
    func testRootsWithSKIArePreferred() async throws {
```

PR #136, opened 2023-09-22
Support loading trust root CAs on Linux

dnadoba:
### Motivation
We want to eventually replace the BoringSSL X.509 layer in `swift-nio-ssl` with `swift-certificates`. One missing feature in `swift-certificates` is the loading of trusted root CAs from the system.

### Changes
- add `CertificateStore.systemTrustRoots` that lazily loads the trusted root certificates from disk.
- add `CertificateStore.insert(_:)` and friends to add additional trusted root CAs to the `CertificateStore.systemTrustRoots` `CertificateStore`
- add `LockedValueBox`, `Promise` and `Future` to the internal API to support lazy loading of the trust roots
- log diagnostics if loading of the system trust roots fails during verification

### Result
`swift-certificates` can efficiently load the installed trusted root CAs on Linux.


Tests/X509Tests/CertificateStore.swift
```diff
@@ -0,0 +1,61 @@
...
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftCertificates project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import XCTest
+import SwiftASN1
+@_spi(Testing)@testable import X509
```
dnadoba, line 17: cc @glbrntt

dnadoba:
5.9 (nightly) and main nightly crashes because of this compiler bug: https://github.com/apple/swift/issues/68708

Sources/X509/Verifier/TrustRootLoading.swift
```diff
@@ -0,0 +1,98 @@
...
+
+extension CertificateStore {
+    /// A ``CertificateStore`` that includes all root Certificate Authorities (CAs) that
+    /// are installed in the systems trust store.
+    ///
+    /// You can add additional trust roots by ``CertificateStore/inserting(_:)-5sc2d`` them into the returned ``CertificateStore``.
+    ///
+    /// - Note: Access this property as early as possible. It will start loading and parsing of the certificates in the background.
+    /// Accessing this property does **not** block.
+    /// - Warning: This property is only available on Linux.
```
Lukasa, line 39: We shouldn't do that: we should make it available everywhere, even if it isn't very useful. Generally we avoid having API only available on one platform. However, we should still have a warning that says that on Apple platforms and on Windows this will not actually find the system trust store, and that the platform verifier should be used instead.

Sources/X509/Verifier/TrustRootLoading.swift
```diff
@@ -0,0 +1,98 @@
...
+        qos: .userInteractive
+    ).asyncFuture {
+        try Self.loadTrustRoot(at: rootCAFileSearchPaths)
+    }
+}
+#endif
+
+extension CertificateStore {
+    @_spi(Testing)
+    public static func loadTrustRoot(at searchPaths: [String]) throws -> CertificateStore {
```
Lukasa, line 57:
```suggestion
    public static func loadTrustRoots(at searchPaths: [String]) throws -> CertificateStore {
```

Tests/X509Tests/CertificateStore.swift
Lukasa, line 17:
```suggestion
@_spi(Testing) @testable import X509
```

Sources/X509/Verifier/CertificateStore.swift
```diff
@@ -12,47 +12,166 @@
...
+    enum Element: Sendable, Hashable {
+        #if os(Linux)
+        case trustRoots
+        #endif
+        /// Stores the certificates, indexed by subject name.
+        case customCertificates([DistinguishedName: [Certificate]])
+    }
+    
+    @usableFromInline
+    var _certificates: _TinyArray<Element>
```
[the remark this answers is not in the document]
Lukasa: I think sticking with the tuple is better.

tayloraswift:
> 5.9 (nightly) and main nightly crashes because of this compiler bug: [apple/swift#68708](https://github.com/apple/swift/issues/68708)

this package seems to be experiencing a second, seemingly different compiler crash as well: https://github.com/apple/swift/issues/68767 , which affects documentation generation.

dnadoba:
@tayloraswift thanks for reporting! The crash you have discovered is even on 5.9-release, mine is only on 5.9 nightly. I have worked around this issue for now. Not sure if we can workaround the other crash.

Sources/X509/Verifier/CertificateStore.swift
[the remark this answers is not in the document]
dnadoba: I have now implemented `_TinyArray2` which store up to two elements inline.

dnadoba:
API break is a false positive:
```
10:02:32   💔 API breakage: constructor CertificateStore.init(_:) has generic signature change from <Certificates where Certificates : Swift.Sequence, Certificates.Element == X509.Certificate> to 
```
We have changed the signature but only from:
```swift
public init<Certificates: Sequence>(_ certificates: Certificates) where Certificates.Element == Certificate
```
to
```swift
public init(_ certificates: some Sequence<Certificate>)
```
which is the same thing, just with a different spelling.

FranzBusch:
@dnadoba let's open an issue on the Swift repo to track this. The API checker should get smart enough to diagnose this correctly.

Sources/X509/PromiseAndFuture.swift
```diff
@@ -0,0 +1,123 @@
...
+                for observer in observers {
+                    observer.resume(returning: result)
+                }
+                state = .fulfilled(result)
+            }
+        }
+    }
+
+    deinit {
+        switch self.state.unsafeUnlockedValue {
```
Lukasa, line 61: I don't think this optimization is appropriate: we can just take the lock, it'll be uncontended.

Sources/X509/Verifier/CertificateStore.swift
```diff
@@ -12,47 +12,133 @@
...
+
 /// A collection of ``Certificate`` objects for use in a verifier.
 public struct CertificateStore: Sendable, Hashable {
-    /// Stores the certificates, indexed by subject name.
+
+    @usableFromInline
+    var systemTrustStore: Bool
     @usableFromInline
-    var _certificates: [DistinguishedName: [Certificate]]
+    var additionTrustRoots: [DistinguishedName: [Certificate]]
```
Lukasa, line 23: Nit: "additional"

Sources/X509/Verifier/CertificateStore.swift
```diff
@@ -12,47 +12,133 @@
...
+
+extension CertificateStore {
+    @usableFromInline
+    struct Resolved {
+
+        @usableFromInline
+        var systemTrustRoots: [DistinguishedName: [Certificate]]
+
+        @usableFromInline
+        var additionTrustRoots: [DistinguishedName: [Certificate]]
```
Lukasa, line 80: Nit: additional.

Sources/X509/Verifier/TrustRootLoading.swift
```diff
@@ -0,0 +1,99 @@
...
+    /// On Darwin based platforms (e.g. macOS, iOS) use Security.framework to validate that a certificate chains up to a trusted root CA.
+    /// On other platforms (e.g. Windows) use the platform verifier.
+    public static let systemTrustRoots: CertificateStore = {
+        // access `cachedTrustRootsFuture` to kick off loading on a background thread
+        _ = cachedSystemTrustRootsFuture
+        return CertificateStore(systemTrustStore: true)
+    }()
+
+    static let cachedSystemTrustRootsFuture: Future<[DistinguishedName: [Certificate]], any Error> =
+        DispatchQueue.global(
```
Lukasa, line 44: Prefer to create a serial dispatch queue for this than directly dispatching to a global queue.

Lukasa (approved):

Lukasa:
Merging over API breakage checker, which is wrong here.

PR #149, opened 2023-10-12
Remove note from README about API instability

yim-lee:
I think we can remove that note now that the library has 1.0.0 tag.


Lukasa (approved):
I agree

PR #150, opened 2023-10-16
Align `swift-format` configuration with other projects

dnadoba:
No code change. We already followed these rules.


PeterAdams-A (approved):
What a brilliant change.

PR #154, opened 2023-12-01
Support converting RDNAttribute values into Strings

Lukasa:
Motivation

RDNAttribute Values are mostly strings, and it's useful to be able to convert them. This patch adds that support.

Modifications

Add support for converting RDN attribute values into Strings.

Result

Easier to work with distinguished names.


dnadoba (approved):
LGTM

PR #155, opened 2023-12-04
Simplify the internal representation of RDNAttribute.Value

Lukasa:
Motivation

The 1.1.0 release included a patch (#154) that supported converting RDNAttribute values into strings. This patch missed that the `.any` case may also have strings in it.

While investigating this limitation, I realised that the constructor function was excessively complex. While it attempted to coerce based on extra information in the ASN.1 specification, it always fell back into the `.any` case. This meant that as a practical matter we always took the same few paths, and so could safely simplify the code.

Modifications

Simplify DER decoding of RDNAttribute.Values.
Introspect the .any cases for UTF8 strings and Printable strings when converting to String.

Result

Faster construction, more flexible representation.

### DIFF
--- a/Sources/X509/RDNAttribute.swift
+++ b/Sources/X509/RDNAttribute.swift
@@ -381,123 +381,11 @@ extension RelativeDistinguishedName.Attribute: DERImplicitlyTaggable {
             }
 
             let value: Value
-            switch type {
-            /// ```
-            /// id-at-commonName        AttributeType ::= { id-at 3 }
-            ///
-            /// -- Naming attributes of type X520CommonName:
-            /// --   X520CommonName ::= DirectoryName (SIZE (1..ub-common-name))
-            /// --
-            /// -- Expanded to avoid parameterized type:
-            /// X520CommonName ::= CHOICE {
-            ///       teletexString     TeletexString   (SIZE (1..ub-common-name)),
-            ///       printableString   PrintableString (SIZE (1..ub-common-name)),
-            ///       universalString   UniversalString (SIZE (1..ub-common-name)),
-            ///       utf8String        UTF8String      (SIZE (1..ub-common-name)),
-            ///       bmpString         BMPString       (SIZE (1..ub-common-name)) }
-            ///
-            ///
-            /// -- Naming attributes of type X520LocalityName
-            ///
-            /// id-at-localityName      AttributeType ::= { id-at 7 }
-            ///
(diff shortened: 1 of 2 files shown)


Sources/X509/RDNAttribute.swift
```diff
@@ -594,8 +482,14 @@ extension String {
...
-        default:
-            return nil
+        case .any(let value):
+            if let printable = try? ASN1PrintableString(asn1Any: value) {
+                self = String(printable)
+            } else if let utf8 = try? ASN1UTF8String(asn1Any: value) {
+                self = String(utf8)
+            } else {
+                return nil
+            }
```
dnadoba, line 492: That's rather inefficient as this will very likely parse twice before returning nil. We should instead parse the data once, and switch over the identifier to decided if it is a UTF8 or Printable String like we do in the initialiser. 

The other option would be that we always "normalise" the data and parse it into  `.utf8String`, `.printableString` or `.any` where `.any` can never contain bytes that would be a valid `.utf8String` or `.printableString`.
This allows us to remove the parsing logic here, but more importantly, it allows us to get rid of the custom Equatable/Hashable conformance and their related types which would be big reduction in code complexity. The only place where we currently don't do this is in `RelativeDistinguishedName.Attribute.Value.init(asn1Any:)`.

Sources/X509/RDNAttribute.swift
```diff
@@ -381,123 +381,11 @@ extension RelativeDistinguishedName.Attribute: DERImplicitlyTaggable {
...
-            /// ```
-            case .RDNAttributeType.countryName:
-                switch valueNode.identifier {
-                case ASN1PrintableString.defaultIdentifier:
-                    value = try .init(printableString: String(ASN1PrintableString(derEncoded: valueNode)))
-                default:
-                    value = .init(storage: .any(ASN1Any(derEncoded: valueNode)))
-                }
-
+            switch valueNode.identifier {
```
dnadoba, line 384: Neat! I didn't know that this was allowed but it makes sense

dnadoba (changes requested):

Sources/X509/RDNAttribute.swift
Lukasa: It's not _generally_ expected as many types don't preserve it, but it's allowed here.

dnadoba:
Closing in favour of https://github.com/apple/swift-certificates/pull/156

PR #156, opened 2023-12-05
Simplify the internal representation of `RDN.Attribute.Value`

dnadoba:
### Motivation

The 1.1.0 release included a patch (#154) that supported converting RDNAttribute values into strings. This patch missed that the `.any` case may also have strings in it.

While investigating this limitation, @Lukasa  realised that the constructor function was excessively complex. While it attempted to coerce based on extra information in the ASN.1 specification, it always fell back into the `.any` case. This meant that as a practical matter we always took the same few paths, and so could safely simplify the code.

### Modifications

- Simplify DER decoding of RDNAttribute.Values
- Normalise internal storage to guarantee that `.any` can never contain bytes which are equal to the DER representation of `.printable` or `.utf8`.
- The above invariant allows us to use the synthesised Hashable implementation which in turn allows to delete the custom `Hashable` implementation 

### Result

Faster construction, normalised representation, less code.

### DIFF
--- a/Sources/X509/RDNAttribute.swift
+++ b/Sources/X509/RDNAttribute.swift
@@ -381,123 +381,11 @@ extension RelativeDistinguishedName.Attribute: DERImplicitlyTaggable {
             }
 
             let value: Value
-            switch type {
-            /// ```
-            /// id-at-commonName        AttributeType ::= { id-at 3 }
-            ///
-            /// -- Naming attributes of type X520CommonName:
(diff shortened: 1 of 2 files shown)


Lukasa (approved):
Nice change, this looks really good to me.

Lukasa:
@dnadoba you have a soundness issue that needs to be resolved.

dnadoba:
Formatting issue that doesn't reproduce locally. 

PR #157, opened 2023-12-05
Move `ExpressibleByArrayLiteral` conformance of `_TinyArray` to the declaring module

dnadoba:
### Motivation

Nightly compilers start complaining about retroactive conformances.

### Modifications

- move conformance to declaring module
- optimise code to not allocate
- add tests

### Result

no warning. optimal generated code for:

```swift
func foo1(int: Int) -> _TinyArray<Int> {
    [int]
}
```

which now compiles down to:
```
output.foo1(int: Swift.Int) -> output._TinyArray<Swift.Int>:
        mov     rax, rdi
        xor     edx, edx
        ret
```
(only with nightly compiler)

https://godbolt.org/z/a91fPe6or

### DIFF
--- a/Sources/X509/CryptographicMessageSyntax/CMSSignature.swift
+++ b/Sources/X509/CryptographicMessageSyntax/CMSSignature.swift
@@ -13,7 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 import SwiftASN1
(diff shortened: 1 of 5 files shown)


Lukasa (approved):
Nice clean patch, thanks @dnadoba

PR #158, opened 2023-12-08
build: add support to vendor dependencies

compnerd:
When building swift-crypto outside of the toolchain (e.g. to build LSP), it is convenient to support vendoring dependencies.  Add support for falling back to local builds of the dependencies when not provided as an existing build.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,7 +64,7 @@ if(NOT SwiftCrypto_FOUND)
   message("-- Vending swift-crypto")
   FetchContent_Declare(Crypto
     GIT_REPOSITORY https://github.com/apple/swift-crypto
-    GIT_TAG 3.0.0)
+    GIT_TAG 3.2.0)
   FetchContent_MakeAvailable(Crypto)
 endif()
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,10 @@
 ##
 ##===----------------------------------------------------------------------===##
 
+if(POLICY CMP0077)
+  cmake_policy(SET CMP0077 NEW)
+endif()
+
 if(POLICY CMP0091)
   cmake_policy(SET CMP0091 NEW)
 endif()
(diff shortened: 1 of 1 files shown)


compnerd:
CC: @yim-lee

CMakeLists.txt
```diff
@@ -37,11 +41,35 @@ set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
...
+
+set(_CERT_SAVED_BUILD_EXAMPLES ${BUILD_EXAMPLES})
+set(_CERT_SAVED_BUILD_TESTING ${BUILD_TESTING})
+
+find_package(SwiftASN1 CONFIG)
+if(NOT SwiftASN1_FOUND)
+  message("-- Vending swift-asn1")
+  FetchContent_Declare(ASN1
+    GIT_REPOSITORY https://github.com/apple/swift-asn1
+    GIT_TAG 1.1.0)
```
yim-lee, line 58: It is highly likely that we will forget to keep this in-sync with `Package.swift`. Would it be better to require users of this feature to set a version environment variable so they don't use a wrong version by accident?

compnerd: That does make it a lot more complicated to use.  The intention here is to enable the use of the CMake based build to debug LSP, which requires SPM, which requires this package.

Lukasa: @yim-lee Can we amend the soundness script to validate that these two match?

yim-lee: If we look at the rule for `swift-crypto`:

```
.package(url: "https://github.com/apple/swift-crypto.git", "2.5.0" ..< "4.0.0"),
```

What's the `swift-crypto` version that we would extract from this for comparison/validation?

If we extract from `Package.resolved` instead, the resolved version may not necessarily match the one in this file. This might be annoying but could work I suppose? I don't think we do `swift build` or `swift package resolve` in the soundness check though.

Lukasa: Yeah, these are all good questions. I'm just wondering if we can get to a local maximum of "least likely to be outside the band of reasonableness".

compnerd:
@yim-lee ping? This has been sitting for a while, and is blocking other work. Can we please merge it and address the feedback for tooling in a follow up?

yim-lee:
@compnerd Frankly I still prefer that we don't hardcode dependency versions, but I don't have time to look into alternatives. If we must merge this PR as-is, then I feel strongly that at the minimum the owners of this repository should not be held responsible for updating the versions. Can we agree to that?

compnerd:
@yim-lee I think we can agree to that - I expect that the occasional patch to bring this back into sync isn't a big deal, but having it track precisely is not a requirement.

CMakeLists.txt
```diff
@@ -37,11 +41,35 @@ set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
...
+  FetchContent_MakeAvailable(ASN1)
+endif()
+
+find_package(SwiftCrypto CONFIG)
+if(NOT SwiftCrypto_FOUND)
+  message("-- Vending swift-crypto")
+  FetchContent_Declare(Crypto
+    GIT_REPOSITORY https://github.com/apple/swift-crypto
+    GIT_TAG 3.0.0)
```
yim-lee, line 67:
```suggestion
    GIT_TAG 3.2.0)
```

yim-lee:

CMakeLists.txt
yim-lee: Do we want these versions to match update-checkout or it doesn't matter?

compnerd: Lets go ahead and match it :)

compnerd:

yim-lee:
@compnerd OK. I have a question about keeping the versions in-sync with update-checkout and whether we want to use the latest, otherwise I can merge this if you think it's ready.

compnerd:
I think that we could use the latest, but I feel more comfortable with trying to match the update-checkout because it is going to be the closest to what is actually in the toolchain. The primary use of this is for ease of debugging projects like the LSP which cannot be built with SPM on Windows due to SPM limitations.

CMakeLists.txt
yim-lee: The problem is update-checkout use swift-asn1 1.0.0 and swift-crypto 3.0.0. lol

But if it's ok to use the latest (i.e., 1.1.0 and 3.2.0) then let's leave it.

yim-lee:
Yeah, update-checkout doesn't use the latest:

https://github.com/apple/swift/blob/main/utils/update_checkout/update-checkout-config.json#L116-L118

compnerd:
BTW, FTR, this is not the mechanism for building the toolchain - the toolchain builds rely on update-checkout and build everything from that checkout, which is why drift here is not a major concern - this is a developer productivity solution.

yim-lee (approved):

PR #166, opened 2024-03-21
Fix `Sendable` warnings from latest `main` compiler

dnadoba:


Lukasa:
Error, womp womp.

Lukasa (approved):

Lukasa:
We now have a compiler crash: we don't need to block the merge on that.

dnadoba:
Compiler crash should be fixed by https://github.com/apple/swift/pull/72433

PR #171, opened 2024-04-22
Add support for validating server hostnames and IP addresses

Lukasa:
Motivation

To make Swift Certificates useful for the Web PKI it's vital that we can validate that a particular certificate is authoritative for a specific domain. This requires a new verifier policy.

Modifications

Added ServerIdentityPolicy
Added tests

Result

We can validate the identity of remote servers.


Sources/X509/Verifier/ServerIdentityPolicy.swift
```diff
@@ -0,0 +1,493 @@
...
+/// A ``VerifierPolicy`` that validates that the leaf certificate is authoritative
+/// for a given hostname or IP address.
+///
+/// This policy is most commonly used to validate the leaf certificate presented by a server
+/// during a TLS handshake.
+///
+/// This policy implements the logic for service validation as specified by
+/// RFC 6125 (https://tools.ietf.org/search/rfc6125), which loosely speaking
+/// defines the common algorithm used for validating that an X.509 certificate
+/// is valid for a given service
```
FranzBusch, line 27: NIT
```suggestion
/// is valid for a given service.
```

Sources/X509/Verifier/ServerIdentityPolicy.swift
```diff
@@ -0,0 +1,493 @@
...
+    var serverIP: LazyIPAddress?
+
+    @usableFromInline
+    var convertedServerHostname: PreparedServerHostname?
+
+    /// Constructs a new ``ServerIdentityPolicy``.
+    ///
+    /// - parameters:
+    ///     - serverHostname: The hostname used to connect to the server.
+    ///     - serverIP: The IP address of the server, if known.
```
FranzBusch, line 43: Presumably you have to pass one or the other or both right? Can we enforce this in the type contract somehow?

Sources/X509/Verifier/ServerIdentityPolicy.swift
```diff
@@ -0,0 +1,493 @@
...
+    }
+}
+
+extension Optional where Wrapped == ServerIdentityPolicy.LazyIPAddress {
+    /// Converts the value from string to one of the IP address cases and
+    /// returns it.
+    ///
+    /// Does nothing if the value is nil.
+    @inlinable
+    mutating func convert() -> ServerIdentityPolicy.IPAddress? {
```
FranzBusch, line 149: NIT: Shouldn't this rather be an init on `ServerIdentityPolicy.IPAdress` that takes a `ServerIdentityPolicy.LazyIPAddress` and on the optional we can just call `flatMap`?

Lukasa: No, because we modify this value as well.

Sources/X509/Verifier/ServerIdentityPolicy.swift
```diff
@@ -0,0 +1,499 @@
...
+
+extension ServerIdentityPolicy: VerifierPolicy {
+    @inlinable
+    public var verifyingCriticalExtensions: [ASN1ObjectIdentifier] {
+        [.X509ExtensionID.subjectAlternativeName]
+    }
+
+    @inlinable
+    public mutating func chainMeetsPolicyRequirements(chain: UnverifiedCertificateChain) -> PolicyEvaluationResult {
+        let targetIP = self.serverIP.convert()
```
dnadoba, line 62: So this is caching the result. Why don't we do this just in the init? Is this laziness actually useful?

Sources/X509/Verifier/ServerIdentityPolicy.swift
```diff
@@ -0,0 +1,499 @@
...
+
+        if let serverHostname = self.serverHostname, self.convertedServerHostname == nil {
+            self.convertedServerHostname = PreparedServerHostname(lowercaseASCIIBytes: serverHostname)
+        }
```
dnadoba, line 66: We are doing another way of caching here as well but forgot to set `serverHostname` to `nil` if the failable  `PreparedServerHostname.init` returns nil. This means we repeatedly are trying to prepare the server hostname if it isn't a valid hostname. I suggest we move this to an init or model the internal state as enum so we are sure we only do initial conversion only once.

Sources/X509/Verifier/ServerIdentityPolicy.swift
Lukasa: The advantage of the laziness here is that if we never have to analyse a cert, we never do the construction.

Sources/X509/Verifier/ServerIdentityPolicy.swift
Lukasa: We can do the enum state, that works for me.

Sources/X509/Verifier/ServerIdentityPolicy.swift
```diff
@@ -0,0 +1,524 @@
...
+extension Collection {
+    /// Splits a collection in two around a given index. This index may be nil, in which case the split
+    /// will occur around the end.
+    fileprivate func splitAroundIndex(_ index: Index?) -> (SubSequence, SubSequence) {
+        guard let index = index else {
+            return (self[...], self[self.endIndex...])
+        }
+
+        let subsequentIndex = self.index(after: index)
+        return (self[..<index], self[subsequentIndex...])
```
dnadoba, line 382: This will crash if `index` is `self.endIndex`

PR #172, opened 2024-04-24
Add OneOfPolicies and AllOfPolicies

hamzahrmalik:
Created a new `OneOfPolicyBuilder`
This is similar to the existing PolicyBuilder
The difference is that the new `OneOfPolicyBuilder` requires only one of the child policies to be valid, not necessarily all of them

A new struct, `OneOfPolicies` implements VerifierPolicy and can be constructed using `OneOfPolicyBuilder`

This can be used within a larger policy e.g.
```swift
let x = Verifier(rootCertificates: .init([])) {
    RFC5280Policy(validationTime: .init()) // this must be met
    OneOfPolicies {
        SomethingPolicy() // one of these 2 must be met
        SomethingElsePolicy() // one of these 2 must be met
    }
}
```

The `verifyingCriticalExtensions` of `OneOfPolicies` is intersection of the `verifyingCriticalExtensions` of each of the child policies, because the overall policy will only be guaranteed to verify an extension if every child policy does so, because we don't know which child policies will actually be evaluated (we stop evaluating when any one passes)

`OneOfPolicies` will only fail if all of the child policies fail. In this case, the failure reason will be a concatenation of all the child failure reasons, with the word 'and' in between them

This PR also introduces `AllOfPolicies`
This is to allow users to specify a set of policies which must always be set, as a single item within a OneOfPolicies


Lukasa:

Sources/X509/Verifier/OneOfPolicyBuilder.swift
```diff
@@ -0,0 +1,207 @@
...
+        mutating func chainMeetsPolicyRequirements(chain: UnverifiedCertificateChain) async -> PolicyEvaluationResult {
+            .failsToMeetPolicy(reason: "No policies specified in OneOfPolicies block")
+        }
+    }
+
+    @inlinable
+    public static func buildBlock() -> some VerifierPolicy {
+        Empty()
+    }
+}
```
dnadoba, line 51: We don't need to support empty OneOfPolicies. We can remove this.

Sources/X509/Verifier/OneOfPolicyBuilder.swift
```diff
@@ -0,0 +1,207 @@
...
+
+    @inlinable
+    public static func buildFinalResult(_ component: AnyPolicy) -> AnyPolicy {
+        func unwrapExistentialAndCache(policy: some VerifierPolicy) -> some VerifierPolicy {
+            PolicyBuilder.CachedVerifyingCriticalExtensions(wrapped: policy)
+        }
+        let cachedPolicy = unwrapExistentialAndCache(policy: component.policy)
+        return AnyPolicy(cachedPolicy)
+    }
+}
```
dnadoba, line 147: Caching is only required at the top level and is done by `PolicyBuilder` already. `OneOfPolicy` will only be used inside a `PolicyBuilder` so we can remove this.

Tests/X509Tests/PolicyBuilderTests.swift
```diff
@@ -328,4 +408,97 @@ final class PolicyBuilderTests: XCTestCase {
...
+        }
+        await assertFailsToMeetPolicy {
+            OneOfPolicies {
+                Policy(result: .failsToMeetPolicy(reason: ""))
+                AllOfPolicies {
+                    Policy(result: .meetsPolicy)
+                    Policy(result: .failsToMeetPolicy(reason: ""))
+                }
+            }
+        }
```
dnadoba, line 502: Can we add tests for `if`, `if` + `else` and `if let` similar to the `PolicyBuilder` tests above?

Sources/X509/Verifier/OneOfPolicyBuilder.swift
```diff
@@ -0,0 +1,207 @@
...
+//===----------------------------------------------------------------------===//
+
+import SwiftASN1
+
+/// Provides a result-builder style DSL for constructing a ``VerifierPolicy`` in which one of the specified policies must match
+///
+/// This DSL allows us to construct dynamic ``VerifierPolicy`` at runtime without using type erasure.
+/// The resulting ``VerifierPolicy`` will use the listed policy in the order of declaration to check if a chain meets any one policy
+@resultBuilder
+public struct OneOfPolicyBuilder {}
```
dnadoba, line 22: We should mention what the DLS is i.e. `OneOfPolicy` with a short code example.

Sources/X509/Verifier/OneOfPolicyBuilder.swift
```diff
@@ -0,0 +1,207 @@
...
+/// let verifier = Verifier(rootCertificates: CertificateStore()) {
+///     RFC5280Policy(validationTime: Date())
+///     OneOfPolicies {
+///         AllOfPolicies {
+///             PolicyA()
+///             PolicyB()
+///         }
+///         PolicyC()
+///     }
+/// }
```
dnadoba, line 191: This should be in a swift code block

Sources/X509/Verifier/OneOfPolicyBuilder.swift
```diff
@@ -0,0 +1,207 @@
...
+public struct AllOfPolicies<Policy: VerifierPolicy>: VerifierPolicy {
```
dnadoba, line 192: Every method needs to be `@inlinable` because we are in a generic declaration.

Sources/X509/Verifier/OneOfPolicyBuilder.swift
```diff
@@ -0,0 +1,207 @@
...
+/// For example, the following policy requires that RFC5280Policy is always met, and either PolicyA or PolicyB is met
+/// It does not require that both PolicyA and PolicyB are met
+/// let verifier = Verifier(rootCertificates: CertificateStore()) {
+///     RFC5280Policy(validationTime: Date())
+///     OneOfPolicies {
+///         PolicyA()
+///         PolicyB()
+///     }
+/// }
+public struct OneOfPolicies<Policy: VerifierPolicy>: VerifierPolicy {
```
dnadoba, line 159: Every method needs to be `@inlinable` because we are in a generic declaration.

Sources/X509/Verifier/OneOfPolicyBuilder.swift
```diff
@@ -0,0 +1,207 @@
...
+/// Use this to build a policy where any one of the sub-policies must be met for the overall policy to be met
+/// For example, the following policy requires that RFC5280Policy is always met, and either PolicyA or PolicyB is met
+/// It does not require that both PolicyA and PolicyB are met
+/// let verifier = Verifier(rootCertificates: CertificateStore()) {
+///     RFC5280Policy(validationTime: Date())
+///     OneOfPolicies {
+///         PolicyA()
+///         PolicyB()
+///     }
+/// }
```
dnadoba, line 158: This should be in a swift code block

dnadoba (changes requested):
Great addition! A shame we can't use parameter packs which :/
I would be great if we could also add a benchmark that uses `OneOfPolicy`. It would be especially interesting to see the difference `@inlinable` annotations make in this case.

Sources/X509/Verifier/OneOfPolicyBuilder.swift
```diff
@@ -0,0 +1,207 @@
...
+    }
+}
+
+// MARK: if
+extension OneOfPolicyBuilder {
+    @inlinable
+    public static func buildOptional(_ component: (some VerifierPolicy)?) -> some VerifierPolicy {
+        PolicyBuilder.WrappedOptional(component)
+    }
+}
```
dnadoba, line 114: I think we need a custom implementation here. With he current implementation this will always meet the policy:
```swift
    Verifier(rootCertificates: .init()) {
        OneOfPolicies {
            if false {
                RFC5280Policy(validationTime: Date())
            }
        }
    }
```
but I think this should fail the policy.

We need a custom WrappedOptional type that failsToMeet the policy if `component` is `nil`.

Sources/X509/Verifier/OneOfPolicyBuilder.swift
dnadoba: We could also keep it and mark it unavailable.
With that being said, I'm actually not feeling to strong here as it has the right behaviour so we can keep it.
Ideally we would also report an error for only one policy because it would be useless.However, it also isn't unsafe and we don't run into any unexpected behavior. It does what it looks like so it's fine that users can write this code after all.

Sources/X509/Verifier/OneOfPolicies.swift
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        @inlinable
+        mutating func chainMeetsPolicyRequirements(chain: UnverifiedCertificateChain) async -> PolicyEvaluationResult {
+            let firstResult = await self.first.chainMeetsPolicyRequirements(chain: chain)
+            switch firstResult {
+            case .meetsPolicy:
+                return .meetsPolicy
+            case .failsToMeetPolicy(let firstReason):
+                let secondResult = await self.second.chainMeetsPolicyRequirements(chain: chain)
```
dnadoba, line 91: hm... it would be great if we could log a diagnostics here but we don't have access to the `diagnosticsHandler` given to `verify`. We could add a new method with the log handler as an "optional" requirement  to the `VerifierPolicy` protocol in an API stable way. Not great but possible. Do you have thoughts on this @Lukasa?

Sources/X509/Verifier/PolicyBuilder.swift
```diff
@@ -111,10 +111,13 @@ extension PolicyBuilder {
     struct WrappedOptional<Wrapped>: VerifierPolicy where Wrapped: VerifierPolicy {
         @usableFromInline
         var wrapped: Wrapped?
+        @usableFromInline
+        var defaultResult: PolicyEvaluationResult
```
dnadoba, line 115: I would prefer two separate implementations that have the default result hardcoded. Policies are already getting really large as they compose into a single inline memory allocation and we don't need this dynamic behaviour.

Sources/X509/Verifier/OneOfPolicies.swift
dnadoba: To be concrete, I'm thinking about something like this:
```swift
extension VerifierPolicy {
    ...

    mutating func chainMeetsPolicyRequirements(chain: UnverifiedCertificateChain, logger: VerifierLogger) async -> PolicyEvaluationResult
}

extension VerifierPolicy {
    public mutating func chainMeetsPolicyRequirements(chain: UnverifiedCertificateChain, logger: VerifierLogger) async -> PolicyEvaluationResult {
        await self.chainMeetsPolicyRequirements(chain: chain)
    }
}

public struct VerifierLogger {
    var _log: Optional<(VerificationDiagnostic) -> ()>
    internal func log(_ diagnostics: @autoclosure @Sendable () -> VerificationDiagnostic) {
        _log?(diagnostics())
    }
    public func log(_ diagnostics: @autoclosure @Sendable () -> String) {
        _log?(/* user defined diagnostics */)
    }
}
```

Sources/X509/Verifier/OneOfPolicyBuilder.swift
hamzahrmalik: talked to @Lukasa offline and we prefer to keep it as is, because there is no harm in having an empty policy

Sources/X509/Verifier/OneOfPolicies.swift
hamzahrmalik: This is a good idea. However, considering the size of this PR already, do you think we could do this in a followup?

dnadoba: Sure

Sources/X509/Verifier/OneOfPolicies.swift
```diff
@@ -0,0 +1,194 @@
...
+
+        @inlinable
+        var verifyingCriticalExtensions: [SwiftASN1.ASN1ObjectIdentifier] {
+            self.wrapped?.verifyingCriticalExtensions ?? []
+        }
+
+        @inlinable
+        mutating func chainMeetsPolicyRequirements(chain: UnverifiedCertificateChain) async -> PolicyEvaluationResult {
+            await self.wrapped?.chainMeetsPolicyRequirements(chain: chain)
+                ?? .failsToMeetPolicy(reason: "No policies specified in OneOfPolicies block")
```
dnadoba, line 136: nit: the message isn't entirely correct. One policy has been specified but is "disable"/"not active". We should include the type of the disabled property as well e.g. `"\(Wrapped.self) in OneOfPolicy is disabled"`

Tests/X509Tests/PolicyBuilderTests.swift
```diff
@@ -328,4 +408,153 @@ final class PolicyBuilderTests: XCTestCase {
...
+
+        let policy: Policy? = nil
+        // This is effectively empty because the optional is always nil
+        await assertFailsToMeetPolicy {
+            OneOfPolicies {
+                if let policy {
+                    policy
+                }
+            }
+        }
```
dnadoba, line 434: Nit: we are missing a test case for a non-nil `WrappedOptional` that meets/fails the policy.

dnadoba (approved):
Two nits and we are good to go

dnadoba (approved):
Great change!

Lukasa (approved):
Love this, thanks!

PR #173, opened 2024-05-07
Allow OneOfPolicies and AllOfPolicies to take throwing policy builders, to match Verifier

hamzahrmalik:
This matches what verifier does (https://github.com/apple/swift-certificates/blob/83640c8097acaec17c9835a083e89678cb0f2b66/Sources/X509/Verifier/Verifier.swift#L22) and allows users to more easily use policies which throw in their initialiser


dnadoba (approved):
LGTM

Lukasa (changes requested):
Can we add a test for this?

FranzBusch (approved):

Lukasa (approved):

PR #175, opened 2024-05-14
Use case-insensitive comparison when validating cert domain name

### LINKED ISSUE #174: Case-insensitive identity verification
[RFC 5280](https://datatracker.ietf.org/doc/html/rfc5280#section-4.2.1.6) states that *while uppercase and lowercase letters are allowed in domain names, no significance is attached to the case*, meaning the case should be ignored when comparing domain names.

Most certificates don't have any uppercase letter in their domain name. But some do.

swift-certificates converts the server hostname to lowercase but doesn't convert the certificate's CN or SAN. Therefore, the domain name comparison always fails when the certificate has uppercase letters in its domain name.

This issue is similar to apple/swift-nio-ssl#463.

baarde:
This PR fixes #174.

### DIFF
--- a/Sources/X509/Verifier/ServerIdentityPolicy.swift
+++ b/Sources/X509/Verifier/ServerIdentityPolicy.swift
@@ -383,6 +383,12 @@ extension Collection {
     }
 }
 
+extension Sequence<UInt8> {
+    fileprivate func caseInsensitiveElementsEqual(_ other: some Sequence<UInt8>) -> Bool {
+        self.elementsEqual(other) { $0.lowercased() == $1.lowercased() }
+    }
+}
+
 extension UInt8 {
     /// Whether this character is a valid DNS character, which is the ASCII
     /// letters, digits, the hypen, and the period.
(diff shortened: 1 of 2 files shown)


Sources/X509/Verifier/ServerIdentityPolicy.swift
```diff
@@ -383,6 +383,12 @@ extension Collection {
     }
 }
 
+extension Sequence<UInt8> {
+    fileprivate func caseInsensitiveElementsEqual(_ other: some Sequence<UInt8>) -> Bool {
+        elementsEqual(other) { $0.lowercased() == $1.lowercased() }
```
Lukasa, line 388: Nit: we prefer explicit `self` here.

Lukasa (changes requested):
Thanks for this, and great catch! The callout is correct, but the citation is wrong. We're implementing the algorithm from RFC 6125, and § 6.4.1 is normative:

> If the DNS domain name portion of a reference identifier is a
   "traditional domain name", then matching of the reference identifier
   against the presented identifier is performed by comparing the set of
   domain name labels using a case-insensitive ASCII comparison

One minor style nit but otherwise this looks great.

Lukasa:

Lukasa (approved):
Beautiful, this is a great patch, thanks.

Lukasa:

Issue #147, opened 2023-10-06
SignedAttributes in CMSSignerInfo


urmat-okonov:
Is it planned to support SignedAttributes and UnsignedAttributes in [CMSSignerInfo](https://github.com/apple/swift-certificates/blob/main/Sources/X509/CryptographicMessageSyntax/CMSSignerInfo.swift)? Our backend requires SignedAttributes to detect SigningTime for validation.

dnadoba:
It is currently not planned but we are open to contributions. `SignedAttributes` and `UnsignedAttributes` are documented in [RFC 5652 - Section 5.3](https://datatracker.ietf.org/doc/html/rfc5652#section-5.3).

PR #176, opened 2024-05-29
Adds support for including signing time in signedAttrs for CMS signing.

R4N:
Motivation:

Many CMS signing validations require the inclusion of signing time in signedAttrs (for example Apple Wallet's Pass CMS signing validation)

Modifications:

* Adds optional includeSigningTime flag to CMS.sign method (defaults to false)
* When includeSigningTime flag is true: includes signing-time (UTCTime ASN1Type), content-type, and message-digest signedAttrs and signs those instead of content bytes (as specified in RFC 5652 section 5.4)
* Adds contentType ASN1ObjectIdentifier
* Adds CMSSigningError (noTimeZone, invalidSigningDateComponent) for when includeSigningTime is true and unable to get UTC time zone or date components
* Adds tests for valid and invalid CMS signature when including signedAttrs

Result:

Providing the includeSigningTime = true flag to CMS.sign will cause signing-time, content-type, and message-digest signedAttrs to be included in the CMS signature and the signature bytes will represent signing those signedAttrs. Using CMS.sign without the flag (defaulting to false) should leave the previous CMS.sign behavior unchanged.

Resolves: https://github.com/apple/swift-certificates/issues/147

Submitted on behalf of Zetetic, LLC

### DIFF
--- a/Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
+++ b/Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
@@ -27,7 +27,7 @@ public enum CMS {
         signingTime: Date? = nil
     ) throws -> [UInt8] {
         if let signingTime = signingTime {
-            return try signWithSigningTime(
+            return try self.signWithSigningTime(
                 bytes,
                 signatureAlgorithm: signatureAlgorithm,
                 certificate: certificate,
@@ -45,7 +45,7 @@ public enum CMS {
             certificate: certificate
         )
 
-        return try serializeSignedData(signedData)
+        return try self.serializeSignedData(signedData)
     }
 
     @inlinable
@@ -93,7 +93,7 @@ public enum CMS {
             certificate: certificate,
             signedAttrs: signedAttrs
         )
-        return try serializeSignedData(signedData)
+        return try self.serializeSignedData(signedData)
     }
 
     @_spi(CMS)
--- a/Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
+++ b/Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
@@ -15,11 +15,6 @@ import Foundation
 import SwiftASN1
 import Crypto
 
-public enum CMSSigningError: Error {
-    case noTimeZone
-    case invalidSigningDateComponent(dateComponents: DateComponents, requiredComponents: Set<Calendar.Component>)
-}
-
 public enum CMS {
     @_spi(CMS)
     @inlinable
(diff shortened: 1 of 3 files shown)


Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
```diff
@@ -23,15 +28,82 @@ public enum CMS {
...
+                Calendar.Component.hour,
+                Calendar.Component.minute,
+                Calendar.Component.second,
+            ]
+            guard let timeZone = TimeZone(identifier: "UTC") else {
+                throw CMSSigningError.noTimeZone
+            }
+            var calendar = Calendar(identifier: .gregorian)
+            calendar.timeZone = timeZone
+            let nowComponents = calendar.dateComponents(desiredComponents, from: Date())
```
Lukasa, line 69: We should avoid this. We have a `Date.utcTime` transformation that already exists.

Lukasa (changes requested):
Thanks for this! I've left a few notes in the diff.

Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
```diff
@@ -23,15 +28,82 @@ public enum CMS {
...
+            // 1. content-type
+            // 2. message-digest
+
+            // add content-type signedAttr cms data
+            let contentTypeVal = try ASN1Any(erasing: ASN1ObjectIdentifier.cmsData)
+            let contentTypeAttribute = CMSAttribute(attrType: .contentType, attrValues: [contentTypeVal])
+            signedAttrs.append(contentTypeAttribute)
+
+            // add message-digest sha256 of provided content bytes
+            try SHA256.hash(data: bytes).withUnsafeBytes { bufferPointer in
```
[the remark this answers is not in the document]
dnadoba: We should not need any unsafe Swift here. The returned digest conforms to `Sequence<UInt8>`. Just convert the returned bytes to an `ArraySlice`.

dnadoba (changes requested):

Tests/X509Tests/CMSTests.swift
```diff
@@ -1014,4 +1055,76 @@ extension CMS {
...
+        // 1. content-type
+        // 2. message-digest
+
+        // add content-type signedAttr cms data
+        let contentTypeVal = try ASN1Any(erasing: ASN1ObjectIdentifier.cmsData)
+        let contentTypeAttribute = CMSAttribute(attrType: .contentType, attrValues: [contentTypeVal])
+        signedAttrs.append(contentTypeAttribute)
+
+        // add message-digest sha256 of provided content bytes
+        try SHA256.hash(data: bytes).withUnsafeBytes { bufferPointer in
```
dnadoba, line 1083: Also, here. No need for unsafe Swift.

R4N:
Thanks for the review/feedback. Definitely good suggestions. Using Date.utcDate to construct the UTCTime from date components allows me to remove the previously defined `CMSSigningError`s as both of the defined errors were related to DateComponents construction.

Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
```diff
@@ -23,15 +23,77 @@ public enum CMS {
         signatureAlgorithm: Certificate.SignatureAlgorithm,
         additionalIntermediateCertificates: [Certificate] = [],
         certificate: Certificate,
-        privateKey: Certificate.PrivateKey
+        privateKey: Certificate.PrivateKey,
+        signingTime: Date? = nil
     ) throws -> [UInt8] {
+        if let signingTime = signingTime {
+            return try signWithSigningTime(
```
Lukasa, line 30: Nit: we prefer to use explicit `self` in all places.

Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
```diff
@@ -23,15 +23,77 @@ public enum CMS {
...
         let signature = try privateKey.sign(bytes: bytes, signatureAlgorithm: signatureAlgorithm)
-        return try sign(
+        let signedData = try self.generateSignedData(
             signatureBytes: ASN1OctetString(signature),
             signatureAlgorithm: signatureAlgorithm,
             additionalIntermediateCertificates: additionalIntermediateCertificates,
             certificate: certificate
         )
+
+        return try serializeSignedData(signedData)
```
Lukasa, line 48: Same nit here, prefer `self.`

Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
```diff
@@ -23,15 +23,77 @@ public enum CMS {
...
+        let signedAttrBytes = coder.serializedBytes[...]
+        let signature = try privateKey.sign(bytes: signedAttrBytes, signatureAlgorithm: signatureAlgorithm)
+        let signedData = try self.generateSignedData(
+            signatureBytes: ASN1OctetString(signature),
+            signatureAlgorithm: signatureAlgorithm,
+            additionalIntermediateCertificates: additionalIntermediateCertificates,
+            certificate: certificate,
+            signedAttrs: signedAttrs
+        )
+        return try serializeSignedData(signedData)
```
Lukasa, line 96: Same note on `self` here.

R4N:
Added in the explicit `self`'s for the internal methods calls within CMSOperations.

Lukasa:

Lukasa:
These failures aren't your fault, our allocation baselines have regressed. I'll update them.

Lukasa:
Feel free to rebase on to the new `main` branch.

R4N:
I believe that should fix it up with the rebase now.

Lukasa:

R4N:
Just rebased off of main for the latest commit that came in. The last `swift-server-bot` check had a failure because of the `CMS.sign` API method signature changing. I believe this is a false positive because of the optional `signingTime` [parameter addition](https://github.com/R4N/swift-certificates/blob/signing-time/Sources/X509/CryptographicMessageSyntax/CMSOperations.swift#L27). Let me know if anything else needs to be done on my end.

Lukasa:

dnadoba (approved):
LGTM. Thanks!

Lukasa:
Merging over the API breakage.

dnadoba:
Note that this is in theory an API breaking change as unapplied methods do not take default values into account:
```swift
// this fails to compile now but previously compiled successfully
let ref = CMS.sign(_:signatureAlgorithm:additionalIntermediateCertificates:certificate:privateKey:)
```
We can partially workaround this limitation by keeping the old signature and introducing an overload with the new parameter. The func with the old signature would just call into the new method. 
However, this only partially solves the source break as
```swift
// error: Ambiguous use of 'CMS.sign'
let ref = CMS.sign
```
now becomes ambiguous and fails to compile as well.

I personally think this is fine though. 
@Lukasa wdyt?

Lukasa:
Yeah I don't mind at all. These are also CMS functions so they're not guaranteed to be API stable, so this isn't very egregious.

PR #178, opened 2024-06-07
Our perf thresholds have changed

Lukasa:


dnadoba (approved):

Lukasa:
The retain/release counts in sync code have become very unstable, and I don't think they're providing us much value in comparison to the time I'm spending trying to dial them in. So I propose we just remove them.

dnadoba:
Fair! If they are not stable, even in sync code, we need to disabled them. It is surprising that this now effects sync code. @hassila are there any known issues?
I'm only aware of https://github.com/ordo-one/package-benchmark/issues/189

Benchmarks/Thresholds/5.10/CertificatesBenchmark.Parse_WebPKI_Roots_from_DER.p90.json
```diff
@@ -1,8 +1,8 @@
 {
-  "mallocCountTotal" : 5506,
+  "mallocCountTotal" : 4851,
   "memoryLeaked" : 0,
   "readSyscalls" : 0,
-  "releaseCount" : 120084,
-  "retainCount" : 108506,
+  "releaseCount" : 120271,
+  "retainCount" : 109425,
```
dnadoba, line 6: Can we remove the retain/release counts from the output as well please?

dnadoba (changes requested):

Benchmarks/Thresholds/5.10/CertificatesBenchmark.Parse_WebPKI_Roots_from_DER.p90.json
Lukasa: Done.

dnadoba (approved):

hassila:
Nothing that I know of that should make them unstable (possibly if you run any sampling metric like thread counts etc? If you run with only ARC metrics enabled is it still unstable?). Then of course it doesn't always add up as mentioned in the related case (some objects are also created with initial ref count which is another factor why we sometime see more releases).

Lukasa:
I don't think we had any sampling metrics enabled: we recorded syscalls, allocations, and retains/releases. And the issue wasn't an adding-up problem: it was that the values changed from run-to-run.

hassila:
Yeah, that's very strange then, that they don't add up is known, but they should be stable.

hassila:
(to be sure to isolate it, you might try running with only arc metrics enabled, then I can't really see how any other parts of the benchmark infrastructure could affect it much)

dnadoba:
@Lukasa have you investigated why the thresholds have changed? How have you updated them?

I'm running `./dev/update-benchmark-thresholds` locally which produces a stable output that I have opened a PR with just now: https://github.com/apple/swift-certificates/pull/184
However, CI doesn't agree on these. These numbers seem to be stable on CI as well.
Docker is used locally and on CI so I wouldn't expect any difference.

Lukasa:
My current theory is that because we aren't scaling the iterations, we're extremely sensitive to minor variations. We should have things settle down if we scale the iteration count by kilo or mega.

Issue #168, opened 2024-04-04
Documentation request: How to create a `Certificate` instance based on `SecCertificateCopyData(:)` DER data?


tscholze:
Would it be possible to add a documentation for the following use case:

Reading a DER-`Data` object via `SecCertificateCopyData(:)` from the keychain and transforming it into a `Certificate` instance to have access to the date values of the X501? I checked the ASN1 documentation but I was not able to find an answer.

Thanks!

Lukasa:
Yes, this would be a really good bit of narrative documentation.

The reality is that it's just two steps: take the object from `SecCertificateCopyData`, pass it to the `Certificate(derEncoded:)` constructor. Everything should work well from there. But it'd be very valuable to add a piece of narrative documentation.

Even nicer might be to add a helper extension into the code that can do this directly. Something like:

```swift
#if canImport(Security)
import Security 

extension SecCertificate {
    public init(_ certificate: Certificate) throws {
        // code here
    }
}

extension Certificate {
    public init(_ certificate: SecCertificate) throws {
        // code here
    }
}
#endif
```

That plus tests would be really nice. Are you interested in adding either?

tscholze:
@Lukasa thanks for your response! I put it on my list :).

I already tried your approach:

```swift

import CryptoKit
import X509
import SwiftASN1

// 1. get the SecCertificate
var rawCert: SecCertificate!
SecIdentityCopyCertificate(identity, &rawCert)
         
// 2. pass it to the constructor   
let certificate = Certificate(derData: SecCertificateCopyData(rawCert))

// 3. Causing an error
'CFData' is not convertible to 'String'
```

Using a `as Data` and than a `String(data:encoding)` was not helpful, too. 
I know this is not a support platform, I just want to understand it - so I can contribute (and get my work done). :)

Lukasa:
Not convertible to String is a weird error. Similarly, `derData` is not one of the constructors we offer.

Does this line work? `let certificate = Certificate(derData: SecCertificateCopyData(rawCert) as Data)`

Lukasa:
Oh sorry, this is the right line:

```
let certificate = Certificate(derEncoded: Array(SecCertificateCopyData(rawCert) as Data))
```

tscholze:
@Lukasa awesome! `derData` was an autocompletion, maybe an hiccup because a property was named the same.

Thank you a lot for providing such an amazing lib. :)

tscholze:
[...]

PR #182, opened 2024-07-02
Add bridge between `Certificate` and `Security.SecCertificate`

aryan-25:
### Motivation:

Currently, there is no way to create an instance of `Certificate` from `Security.SecCertificate` and vice versa (GitHub Issue #168).

### Modifications:

- Added an initializer to `Certificate` which constructs an instance from `Security.SecCertificate`.
- Added a static function `createWithCertificate(_ certificate: Certificate)` in an extension to `Security.SecCertificate` which returns a `SecCertificate` instance.
  - Various errors when trying to add an initializer in an extension of `Security.SecCertificate` -- this is presumably because `SecCertificate` is defined in C.

### Result:

Users can create an instance of `Certificate` from `Security.SecCertificate` and vice versa.

### DIFF
--- a/Sources/X509/Certificate.swift
+++ b/Sources/X509/Certificate.swift
@@ -295,3 +295,28 @@ extension Certificate: PEMRepresentable {
     @inlinable
     public static var defaultPEMDiscriminator: String { "CERTIFICATE" }
 }
+
+#if canImport(Security)
+import Security
+extension Certificate {
+    /// Creates an instance of `Certifcate` from `SecCertificate`
+    /// - Parameter certificate: The `SecCertificate` instance used to initialize this new `Certificate` instance
+    public init(_ certificate: SecCertificate) throws {
+        try self.init(derEncoded: Array(SecCertificateCopyData(certificate) as Data))
+    }
+}
+
+extension SecCertificate {
+    /// Creates an instance of `SecCertificate` from `Certificate`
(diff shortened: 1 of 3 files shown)


Sources/X509/Certificate.swift
```diff
@@ -295,3 +295,28 @@ extension Certificate: PEMRepresentable {
...
+    public init(_ certificate: SecCertificate) throws {
+        try self.init(derEncoded: Array(SecCertificateCopyData(certificate) as Data))
+    }
+}
+
+extension SecCertificate {
+    /// Creates an instance of `SecCertificate` from `Certificate`
+    /// - Parameter certificate: The `Certificate` instance used to initialize this new `SecCertificate` instance
+    /// - Returns: A new `SecCertificate` instance based on the provided `Certificate` instance
+    public static func createWithCertificate(_ certificate: Certificate) throws -> SecCertificate {
```
dnadoba, line 313: This func should have `make` as a prefix instead of create:
> Begin names of factory methods with “make”, e.g. x.makeIterator().
> https://www.swift.org/documentation/api-design-guidelines/#strive-for-fluent-usage

dnadoba (changes requested):
Amazing! This functionality is usually rather difficult to discover. Can we also add a section to the documentation overview that briefly mentions these methods?

dnadoba (approved):

dnadoba:

PR #184, opened 2024-08-10
Update benchmarks by running `./dev/update-benchmark-thresholds`

dnadoba:


Lukasa:
Let's kill two birds with one stone here and move to kilo or mega scaling values on all of these.

Lukasa:
Closed in favour of #189

PR #187, opened 2024-08-14
Import crypto with preconcurrency flag

PeterAdams-A:
Motivation:

Crypto does not yet support concurrency.

Modifications:

Change Crypto imports where necessary to import
with preconcurrency flag.

Result:

Fewer warnings in swift 6.


Lukasa (approved):
LGTM, thanks.

PR #189, opened 2024-08-14
Up our scaling factor to kilo

Lukasa:
This should help with our thresholds moving around so much.


FranzBusch (approved):

dnadoba:
This will likely be too slow

Lukasa:
Yeah, that's fine though, I wanted to see how far we could push it. I've now dropped us back down to kilo.

PR #192, opened 2024-09-06
[CI] Adopt GitHub actions

FranzBusch:


Lukasa:

PR #193, opened 2024-09-06
[CI] Adopt GitHub actions

FranzBusch:


FranzBusch:
@Lukasa @dnadoba any idea why the benchmarks regressed on the nightlies here?

Lukasa:
I'm afraid I don't. It'd be good if someone could find the time to investigate that.

FranzBusch (commented):
This PR is ready for review now @Lukasa. Some of the benchmarks regressed quite a lot on Swift 6 and I will investigate that in a follow up but I want to unblock the CI transition now. If the changes look good to you then please go ahead and disable the old CI and add the new checks as branch protection.

Lukasa (approved):

PR #195, opened 2024-09-09
Support SecKey Certificate private keys

robotrory:
## Motivation

I would like to create `Certificates` using a [`SecKey`](https://developer.apple.com/documentation/security/seckey) so that I can create certificates signed by hardware-backed keys. Currently, Swift Certificates requires the private key be provided in an exported representation, but this isn't possible when a `SecKey` is marked as non-exportable.

## Modifications

Adds a new `SecKeyWrapper` backing to `Certificate.PrivateKey` to support initalisation with `SecKey`. This wrapper struct handles key validation and signing operations by calling the appropriate `SecKey` interfaces. Adds additional tests to support new backing type.

## Result

`Certifcate.PrivateKey` can be initialised with a `SecKey`, allowing existing hardware-backed `SecKey` instances to be used as the issuer private key for new `Certificate` instances.


Lukasa:

Sources/X509/CMakeLists.txt
```diff
@@ -13,105 +13,7 @@
...
-  "Verifier/VerifierPolicy.swift"
-  "X509BaseTypes/AlgorithmIdentifier.swift"
-  "X509BaseTypes/ECDSASignature.swift"
-  "X509BaseTypes/RSAPKCS1PublicKey.swift"
-  "X509BaseTypes/SubjectPublicKeyInfo.swift"
-  "X509BaseTypes/TBSCertificate.swift"
-  "X509BaseTypes/Time.swift"
-  "X509BaseTypes/TimeCalculations.swift"
-  "X509BaseTypes/Validity.swift")
+)
```
Lukasa, line 16: These changes look wrong, want to re-run the script?

robotrory: Thanks for catching this. I was missing `gfind` via `brew install findutils` that was causing the script to fail.

Lukasa (approved):

PR #204, opened 2024-10-15
Avoid the macOS requirement.

### LINKED ISSUE #202: 'kSecAttrKeyTypeECDSA' is unavailable in iOS
I'm getting a build error with the new 1.6 version:
'kSecAttrKeyTypeECDSA' is unavailable in iOS

https://github.com/apple/swift-certificates/blob/d8799da503be1dd95cb9128478b5c3b22451a4b2/Sources/X509/SecKeyWrapper.swift#L122

Lukasa:
It seems that `kSecAttrKeyTypeECDSA` is not available on non-mac platforms. This patch amends the code to run appropriately on those platforms.

I've validated this works by testing in the iOS simulator. I'll need to validate the other platforms as well, but that'll take a little while longer, so I'll make a follow-up patch if this turns out to be insufficient.

Resolves #202.

### DIFF
--- a/Sources/X509/SecKeyWrapper.swift
+++ b/Sources/X509/SecKeyWrapper.swift
@@ -119,7 +119,7 @@ extension Certificate.PrivateKey {
                 )
             }
 
-            if privateKeyType == (kSecAttrKeyTypeECDSA as String) {
+            if privateKeyType == (kSecAttrKeyTypeECSECPrimeRandom as String) {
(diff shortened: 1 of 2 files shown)


FranzBusch (approved):

Lukasa:
I've validated this works fine on iOS, tvOS, and watchOS. visionOS hit a separate bug in my xcode that I'm following up with the compiler team on, but I have no reason to assume it isn't working there too.

PR #205, opened 2024-10-29
Add support for Ed25519 certificates

Lukasa:
Motivation

While Ed25519 is not available in the WebPKI, there are a number of contexts where it's an appropriate signature algorithm for X.509. In those contexts,
we should allow the signature algorithm.

Modifications

Add support for Ed25519 keys and signatures.
Add unit tests

Result

Ed25519 support in the X509 certs

### DIFF
--- a/Sources/_CertificateInternals/CMakeLists.txt
+++ b/Sources/_CertificateInternals/CMakeLists.txt
@@ -13,7 +13,7 @@
 ##===----------------------------------------------------------------------===##
 
 add_library(_CertificateInternals
-  "_TinyArray.swift")
+)
 
 set_target_properties(_CertificateInternals PROPERTIES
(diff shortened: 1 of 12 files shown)


FranzBusch (approved):

Sources/_CertificateInternals/CMakeLists.txt
```diff
@@ -13,7 +13,7 @@
 ##===----------------------------------------------------------------------===##
 
 add_library(_CertificateInternals
-  "_TinyArray.swift")
+)
```
3405691582, line 16: Is this change intended? I get a cmake error now this is a library without any sources.

Lukasa: Hrm, that looks incorrect but CI should have caught it.

Issue #201, opened 2024-10-14
Add `detachedContent: Bool` capability to `CMS.sign` method


lorenzofiamingo:
Maybe I'm missing this capability, but could be added `detachedContent: Bool` (like with [`CMSEncoderSetHasDetachedContent`](https://developer.apple.com/documentation/security/cmsencodersethasdetachedcontent(_:_:)) or [`CMSEncodeContent`](https://developer.apple.com/documentation/security/cmsencodecontent(_:_:_:_:_:_:_:_:))) to [`CMS.sign`](https://github.com/apple/swift-certificates/blob/d8799da503be1dd95cb9128478b5c3b22451a4b2/Sources/X509/CryptographicMessageSyntax/CMSOperations.swift#L21) method?
Thank you for this awesome library!

Lukasa:
Hi @lorenzofiamingo, thanks for filing this issue!

What are you looking to achieve? Currently the library _always_ produces detached signatures. Are you looking to add support for producing attached ones?

lorenzofiamingo:
Hi there! Yes, please do produce the attached one. Thanks you very much!

Lukasa:
Ok, fantastic. This is not super high on my priority list so I won't get to it very soon, but I'll mark this as a good first issue for the community, as I'm happy to walk someone through the changes that would have to be made.

Lukasa:
Thanks for stepping in @jonct! Let's discuss the solution there on the diff.

PR #206, opened 2024-11-04
Add support for attached signatures (#201)

jonct:
Initial rough cut.


Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
```diff
@@ -38,12 +39,17 @@ public enum CMS {
 
         // no signing time provided, sign regularly (without signedAttrs)
         let signature = try privateKey.sign(bytes: bytes, signatureAlgorithm: signatureAlgorithm)
-        let signedData = try self.generateSignedData(
+        var signedData = try self.generateSignedData(
```
Lukasa, line 42: I think we should sink the generation of the signed data down into `generateSignedData`. This has two advantages.

First, it means the extension to the "with signing time" case is straightforward. That case also ends up calling into generateSignedData`, which makes our lives a lot easier.

Secondly, it means we don't have to do the transformation back and forth through `CMSContentInfo`, which makes the operation _much_ cheaper. Transforming through ASN1Any is fairly expensive and it'd be better if we set this up correctly the first time, instead of ping-ponging through.

Lukasa (commented):
Thanks, this is a good start! I've left some notes in the diff with suggestions. I'd also suggest writing some tests, which will force you to also handle the verification path. Fortunately, the verification path is straightforward: accept a `nil` data object and grab the bytes to be verified from the CMS object instead in that case.

Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
Lukasa: So this is broadly looking pretty good.

For test cases, you want to add a few new ones, rather than tweak the existing ones. You probably want the following:

1. A test case that tests that the non-signing-time path can both produce a signature, and have that signature verified, with the signature in non-detached mode. Roughly equivalent to `testSimpleSigningVerifying` but with non-detached signatures.
2. A test case like the above, but with signing time included. Roughly equivalent to `testSigningWithSigningTimeSignedAttr`.
3. A version of (1) where the signature is made non-attached, but is verified as if it were attached. This should fail.
4. A version of (2) where the signature is made non-attached, but is verified as if it were attached. This should fail.
5. 3. A version of (1) where the signature is made attached, but is verified as if it were non-attached. This should pass: all th eneeded data is present.
4. A version of (2) where the signature is made attached, but is verified as if it were nonattached. This should pass: all the needed data is present..

Lukasa: There's absolutely nothing to apologise for, it's not always easy to get the time management lined up. This is the nice thing about OSS, I can be event-driven, so waiting for you to come back isn't a burden at all! I'll re-review now.

Lukasa (approved):
Fantastic, this is really nice work. Thanks @jonct! :sparkles:

Lukasa:
Merging over the API breakage, which is expected.

PR #213, opened 2024-12-13
Enable strict concurrency

clintonpi:
Motivation:

To catch potential data races at build time.

Modifications:

- Bump Swift tools version from 5.8 to 5.9 in Package.swift.
- Enable strict concurrency in Package.swift.
- Adjust the documentation section on the supported Swift versions.
- Implement minor fixes for the surfaced strict concurrency warnings.
- Add `-require-explicit-sendable` to the 6.0, nightly 6.0 and nightly main CI checks.

Result:

Strict concurrency adoption.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -12,7 +12,7 @@ jobs:
     uses: apple/swift-nio/.github/workflows/unit_tests.yml@main
     with:
       linux_5_9_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
-      linux_5_10_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
+      linux_5_10_arguments_override: "-Xswiftc --explicit-target-dependency-import-check error"
       linux_6_0_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error -Xswiftc -require-explicit-sendable"
       linux_nightly_6_0_arguments_override: "--explicit-target-dependency-import-check error -Xswiftc -require-explicit-sendable"
       linux_nightly_main_arguments_override: "--explicit-target-dependency-import-check error -Xswiftc -require-explicit-sendable"
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -13,9 +13,9 @@ jobs:
     with:
       linux_5_9_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
       linux_5_10_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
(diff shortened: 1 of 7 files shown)


Tests/X509Tests/SignatureTests.swift
```diff
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-import XCTest
+@preconcurrency import XCTest
```
Lukasa, line 15: I'm not convinced this is right. What errors did this silence?

Tests/X509Tests/SecKeyWrapperTests.swift
```diff
@@ -12,7 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
-import XCTest
+@preconcurrency import XCTest
```
Lukasa, line 15: Similarly I'm not convinced this is right, what errors did this silence?

Tests/X509Tests/SignatureTests.swift
clintonpi: These:

`Static property 'secKeyRSA' is not concurrency-safe because non-'Sendable' type 'SecKey?' may have shared mutable state; this is an error in the Swift 6 language mode`

Lukasa: Ah, we have a missing import. Instead of putting `@preconcurrency` here, please add this:

```
#if canImport(Darwin)
@preconcurrency import Security
#endif
```

.github/workflows/main.yml
```diff
@@ -13,9 +13,9 @@ jobs:
     with:
       linux_5_9_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
       linux_5_10_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
```
Lukasa, line 15: We have to lose warnings-as-errors on 5.10, as we can no longer maintain it with this change.

.github/workflows/main.yml
```diff
@@ -12,10 +12,10 @@ jobs:
     uses: apple/swift-nio/.github/workflows/unit_tests.yml@main
     with:
       linux_5_9_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
-      linux_5_10_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
-      linux_6_0_arguments_override: "-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error"
-      linux_nightly_6_0_arguments_override: "--explicit-target-dependency-import-check error"
-      linux_nightly_main_arguments_override: "--explicit-target-dependency-import-check error"
+      linux_5_10_arguments_override: "-Xswiftc --explicit-target-dependency-import-check error"
```
Lukasa, line 15: Missed removing the `-Xswiftc`

clintonpi: Oh, thank you!

Lukasa (approved):

PR #214, opened 2024-12-17
Support certificates with invalid RDN attribute values

baarde:
Some certificate authorities issue certificates with invalid RDN attribute values. Certificates must use UTF8String instead of PrintableString to represent strings that contains `*` or `@`, but mistakes happen.

When a RDN attribute value is a PrintableString or UTF8String with an invalid content, rather than rejecting the certificate, fall back to storing the value as an Any.

## Rationale

Certificate authorities using the wrong string type is a common mistake. For example, DigiCert issued intermediate certificates with invalid ampersands to Wells Fargo & Company (golang/go#22970). 

This PR adds *partial* support for certificates that contain invalid PrintableString or UTF8String (less likely) values. When failing to parse a PrintableString or UTF8String, the raw bytes are stored as an Any (same as IA5String, BMPString, ...).

Though those certificates are now parsed without any error, they aren't fully supported yet. Trying to convert the value to a String returns `nil`, so the CN doesn't match. I don't know whether we should add support for converting invalid PrintableString values (and IA5String/BMPString/...) to String in the future. In the meantime, developers who want to read the CN have access the raw Any value.

### DIFF
--- a/Sources/X509/RDNAttribute.swift
+++ b/Sources/X509/RDNAttribute.swift
@@ -119,12 +119,16 @@ extension RelativeDistinguishedName.Attribute.Value {
 extension RelativeDistinguishedName.Attribute.Value.Storage: DERParseable, DERSerializable {
     @inlinable
     init(derEncoded node: SwiftASN1.ASN1Node) throws {
-        switch node.identifier {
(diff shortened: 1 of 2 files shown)


Lukasa (approved):
Nice one, thanks @baarde!

PR #215, opened 2024-12-17
Make RDN.Attribute.Value conversion to ASN1Any public

baarde:
When the value is not a printable or UTF8 string, it may be useful to be able to access the underlying encoded bytes.


baarde:
I think it makes sense to expose this method, as we already have public methods for converting between `Value` and `String` and from `ASN1Any` to `Value` (but not the other way around).

Currently, we can access the value's serialized bytes by encoding then decoding the whole attribute as a custom type. It works but it's a little bit involved.

erikgadireddi:
@baarde i would also appreciate this change. 
I also find out that  RDN.Attribute.Value might not be UTF8 or printable. 
For example teletext string 😅 
So this change makes it easier for me too. 🤞

Lukasa (approved):
This is a good change, and it costs us absolutely nothing. Thanks @baarde!

Issue #203, opened 2024-10-15
CMS signer identifier is incorrectly encoded


Lukasa:
We have the following code for serializing and deserializing the `subjectKeyIdentifier` option in `CMSSignerIdentifier`:

https://github.com/apple/swift-certificates/blob/d8799da503be1dd95cb9128478b5c3b22451a4b2/Sources/X509/CryptographicMessageSyntax/CMSSignerIdentifier.swift#L58-L62

https://github.com/apple/swift-certificates/blob/d8799da503be1dd95cb9128478b5c3b22451a4b2/Sources/X509/CryptographicMessageSyntax/CMSSignerIdentifier.swift#L39-L45

Both of these perform explicit tagging of the field. That's informed by [this declaration from RFC 5652 § 12.1](https://datatracker.ietf.org/doc/html/rfc5652#section-12.1):

```
SignerIdentifier ::= CHOICE {
  issuerAndSerialNumber IssuerAndSerialNumber,
  subjectKeyIdentifier [0] SubjectKeyIdentifier }
```

However, this misses that the tags default for the document is specified as:

```
   DEFINITIONS IMPLICIT TAGS ::=
```

So the default tag is implicit unless one of the exceptions from [ITU 680 § 30.6](https://www.itu.int/ITU-T/studygroups/com17/languages/X.680-0207.pdf) applies. None of them do, so this should be implicitly tagged.

PR #218, opened 2025-01-13
Tag CMS signer identifier implicitly

nicholaskim94:
As pointed out in https://github.com/apple/swift-certificates/issues/203, the SignerIdentifier in PKCS#7 CMS should be implicitly tagged while the currently swift-certificates explicitly tags it. This PR contains changes for SignerIdentifier of SubjectKeyIdentifier type to be implicitly tagged.

[RFC 5652 § 12.1](https://datatracker.ietf.org/doc/html/rfc5652#section-12.1):
```
SignerIdentifier ::= CHOICE {
  issuerAndSerialNumber IssuerAndSerialNumber,
  subjectKeyIdentifier [0] SubjectKeyIdentifier }
  ```
  
  
  Resolves: https://github.com/apple/swift-certificates/issues/203

### DIFF
--- a/Sources/X509/CryptographicMessageSyntax/CMSSignerIdentifier.swift
+++ b/Sources/X509/CryptographicMessageSyntax/CMSSignerIdentifier.swift
@@ -22,7 +22,6 @@ import SwiftASN1
 ///  ```
 @usableFromInline
 enum CMSSignerIdentifier: DERParseable, BERParseable, DERSerializable, BERSerializable, Hashable, Sendable {
-
     @usableFromInline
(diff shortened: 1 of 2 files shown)


Lukasa:
Looks like a little formatting needs to be cleaned up, but otherwise this patch looks great!

nicholaskim94:
@Lukasa Thanks for the prompt review! I've cleaned up the formatting

Lukasa (approved):
Fantastic patch @nicholaskim94, thanks so much! :sparkles:

PR #219, opened 2025-01-13
Use proper hash when computing message digest in signed attributes

baarde:
Per [RFC 5652 § 11.2](https://datatracker.ietf.org/doc/html/rfc5652#section-11.2):

"For signed-data, the message digest is computed using the signer's message digest algorithm."

Currently, when computing the signed attributes' message digest, `CMS.sign` assumes the digest algorithm is SHA256, which is incorrect. The behavior of `CMS.isValid` is already correct and doesn't need updating. I've made `Digest` conform to `Sequence` to facilitate the conversion/comparison of the message digest.

### DIFF
--- a/Sources/X509/Digests.swift
+++ b/Sources/X509/Digests.swift
@@ -43,6 +43,22 @@ enum Digest {
     }
 }
 
+extension Digest: Sequence {
+    @usableFromInline
+    func makeIterator() -> Array<UInt8>.Iterator {
+        switch self {
+        case .insecureSHA1(let sha1):
+            return sha1.makeIterator()
+        case .sha256(let sha256):
+            return sha256.makeIterator()
+        case .sha384(let sha384):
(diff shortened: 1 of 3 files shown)


Sources/X509/Digests.swift
```diff
@@ -43,6 +43,22 @@ enum Digest {
     }
 }
 
+extension Digest: Sequence {
+    @usableFromInline
+    func makeIterator() -> Array<UInt8>.Iterator {
```
Lukasa, line 48: It might be nicer to return this using `some IteratorProtocol`, potentially with the extra `associatedtype` to get that to work. That adds some defense in-case we accidentally make this public later and we regret exposing the array iterator directly.

baarde: Good idea.

Lukasa (approved):
Nice one, this looks really good. Thanks @baarde!

PR #221, opened 2025-01-15
Add default value for signature algorithm

baarde:
Most of the time, the signature algorithm that should be used is dictated by the type of the private key.

* Ed25519 keys support only one signature algorithm.
* RFC 5753 section 8 recommends that "[the P-256 curve] be used with SHA-256; the P-384 curve be used with SHA-384; and the P-521 curve be used with SHA-512".
* RSA keys support 4 signature algorithms. But most people use RSA with SHA-256 and nobody should use RSA with SHA-1 anymore.

More over, Certificate.PrivateKey is opaque to the user, who may not know what type of private key they're using and what the appropriate signature algorithm is.

For those reasons, we add convenience wrappers around methods with a signature algorithm to provide a reasonable default value.


Sources/X509/Certificate.swift
```diff
@@ -209,6 +209,51 @@ public struct Certificate {
         self.signatureBytes = try DER.Serializer.serialized(element: ASN1BitString(self.signature))[...]
     }
 
+    /// Construct a certificate from constituent parts, signed by an issuer key.
+    ///
+    /// This API can be used to construct a ``Certificate`` directly, without an intermediary
+    /// Certificate Signing Request. The ``signature-swift.property`` for this certificate will be produced
+    /// automatically, using `issuerPrivateKey`.
```
Lukasa, line 216: Let's add a section to the doc comment here that explicitly compares this to the other initializer, indicating that it defaults the `signatureAlgorithm` property based on the type of the key.

Sources/X509/CSR/CertificateSigningRequest.swift
```diff
@@ -145,6 +145,32 @@ public struct CertificateSigningRequest {
         self.signatureBytes = try DER.Serializer.serialized(element: ASN1BitString(self.signature))[...]
     }
 
+    /// Construct a CSR for a specific private key.
+    ///
+    /// This API can be used to construct a certificate signing request that can be passed to a certificate
+    /// authority. It will correctly generate a signature over the request.
```
Lukasa, line 151: Let's add a note to point to the other initializer and indicate that this defaults the signature algorithm based on the private key.

Tests/X509Tests/CertificateTests.swift
```diff
@@ -729,4 +729,49 @@ final class CertificateTests: XCTestCase {
...
+            serialNumber: .init(bytes: [1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12]),
+            publicKey: privateKey.publicKey,
+            notValidBefore: Date(),
+            notValidAfter: Date() + 3600,
+            issuer: name,
+            subject: name,
+            extensions: Certificate.Extensions {},
+            issuerPrivateKey: privateKey
+        )
+    }
```
Lukasa, line 776: I know it's a bit of a chore, but can we get equivalent tests for the CMS and CSR use-cases as well?

Lukasa (commented):
Thanks for this! The patch is generally really good, I just left a few doc comment notes and an ask for more tests.

Tests/X509Tests/CertificateTests.swift
baarde: It's done. One test is currently failing, but that should be addressed by #220.

Sources/X509/Certificate.swift
baarde: Good idea.

Sources/X509/CSR/CertificateSigningRequest.swift
baarde: Good idea.

Lukasa (approved):
Very nice, this looks great.

Lukasa:
Looks like there's a few format issues to fix up.

baarde:
OK, I'd finally added `swift fomat lint` as a pre-commit hook. 😅

Lukasa:
Looks like we have some test failures here.

baarde:
That's the problem of putting my commits in independent PRs: CMS signature with Ed25519 doesn't work (yet). 😅 

I've rebased on top of #220. This fixes the tests.

PR #222, opened 2025-01-19
Fix a small error in AlgorithmIdentifier description and add missing cases

baarde:
This PR adds missing cases and fixes a small typo.

I haven't added tests, as `AlgorithmIdentifier` is internal API.


Lukasa:
Thanks a bunch!

PR #239, opened 2025-03-29
Make the usual pthread changes for OpenBSD.

3405691582:


dnadoba (approved):
LGTM. I'm have thought a bit about if we can move to `Mutex` from the standard library but this will likely never be fully possible without a semver major because Mutex is only available on macOS 15.0+ and aligned releases but we support 10.15 :/ 

However, we could only move to Mutex on new platforms like OpenBSD. That means we would need to use our own lock wrapper on non-darwin platforms but could use Mutex on all other platforms. This is somewhat desirable as we would get support for any new platform that the standard library supports for free. 

As Mutex is only available in Swift 6+ toolchains we would also need to add some more #if's to actually make this work. 
Not sure if it is worth the effort while we still support non Swift 6 toolchains.

Lukasa (approved):
LGTM, thanks.

PR #252, opened 2025-05-14
CMake: Use full module triples

etcwilde:
Teach the CMake build to install swiftmodules using the full module triple. This uses the compiler to determine the appropriate module triple instead of attempting to compute the architecture in CMake script, making the build more robust for new platforms and architectures.

Also removing redundant policies. CMP0077 and CMP0091 were added before the minimum CMake version and therefore will always be active in this project.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,14 +12,6 @@
 ##
 ##===----------------------------------------------------------------------===##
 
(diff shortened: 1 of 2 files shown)


glbrntt (approved):
Thanks @etcwilde!

PR #253, opened 2025-05-20
Fix pthread use on FreeBSD

etcwilde:
Much like OpenBSD, the FreeBSD pthread APIs are not imported with the appropriate nullability annotations, so the base pthread mutex type should be an optional value. Eventually we should provide API notes to fix this issue, but that is not in place today.
See https://github.com/swiftlang/swift/issues/81407


glbrntt (approved):
Thanks @etcwilde!

PR #256, opened 2025-06-09
Custom certificate store

brimworks:
This change adds support for implementing a `CustomCertificateStore` in a backwards compatible way with minimal impact to performance. A custom certificate store can be implemented, and then the new `CertificateStore(custom:)` constructor can be used so that the custom certificate store can be used interchangeably with a "normal" certificate store.

Additionally, a test case verifies that this new protocol will work for our use case. More specifically, the use case we have is for normalization of distinguished names so that an exact match of distinguished names is no longer required. For example, if `printableString` should be used interchangeably with a `utf8String`.

This pull request has two commits:
- One that adds support for a `CustomCertificateStore`
- One that modifies the implementation to support `async` lookup of certificiates.

The second change is not required for our use-case, however I would expect that there may be a future use-case that requires looking up certificates in a database, and supporting an `async` protocol would be valuable. Also, the `verify()` method is already async, so supporting an `async` lookup does not disrupt any existing public APIs.

### DIFF
--- a/Sources/X509/CMakeLists.txt
+++ b/Sources/X509/CMakeLists.txt
@@ -92,6 +92,7 @@ add_library(X509
   "Verifier/AllOfPolicies.swift"
   "Verifier/AnyPolicy.swift"
   "Verifier/CertificateStore.swift"
+  "Verifier/CustomCertificateStore.swift"
   "Verifier/OneOfPolicies.swift"
   "Verifier/PolicyBuilder.swift"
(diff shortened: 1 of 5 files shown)


Lukasa (commented):
Awesome, this looks really good. Left a small nit in the diff.

Lukasa (approved):
Thanks for this @brimworks, and sorry for the belated review. We can ignore the CI failures: these are known breakages on `main`.

PR #257, opened 2025-06-23
Support singing via Certificate PrivateKey

bauer-andreas:
This PR introduces two key changes to improve flexibility when working with signatures in swift-certificates. Previously, it was only possible to verify signatures using `Certificate.PrivateKey`, but not to generate them.

- The internal `sign(bytes:signatureAlgorithm:)` method on `Certificate.PrivateKey` is now public, allowing users to generate signatures directly.
- `Certificate.Signature` gains a new `rawRepresentation` property that exposes the raw byte representation of a signature.

### DIFF
--- a/Sources/X509/Signature.swift
+++ b/Sources/X509/Signature.swift
@@ -132,18 +132,9 @@ extension Certificate.Signature {
 extension Certificate.Signature {
     /// The raw byte representation of the signature.
     @inlinable
-    public var rawRepresentation: [UInt8] {
-        switch backing {
-        case let .ecdsa(signature):
-            var serializer = DER.Serializer()
-            // we are serializing from raw bytes, force try is safe here
-            try! signature.serialize(into: &serializer, withIdentifier: ECDSASignature.defaultIdentifier)
-            return serializer.serializedBytes
-        case let .ed25519(data):
-            return .init(data)
-        case let .rsa(signature):
-            return .init(signature.rawRepresentation)
-        }
+    public var rawRepresentation: ArraySlice<UInt8> {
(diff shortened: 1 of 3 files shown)


Sources/X509/Signature.swift
```diff
@@ -128,6 +128,25 @@ extension Certificate.Signature {
     }
 }
 
+@available(macOS 10.15, iOS 13, watchOS 6, tvOS 13, macCatalyst 13, visionOS 1.0, *)
+extension Certificate.Signature {
+    /// The raw byte representation of the signature.
+    @inlinable
+    public var rawRepresentation: [UInt8] {
```
Lukasa, line 135: Given that the ECDSA signatures are being serialized as DER, I think this is more properly `derRepresentation`.

Tests/X509Tests/SignatureTests.swift
```diff
@@ -50,6 +50,28 @@ final class SignatureTests: XCTestCase {
...
+        let input = Array("Hello World".utf8)
+        let privateKey = Certificate.PrivateKey(Self.rsaKey)
+
+        let expected = try Self.rsaKey.signature(for: SHA256.hash(data: input), padding: .insecurePKCS1v1_5)
+        let found = try privateKey.sign(bytes: input, signatureAlgorithm: .sha256WithRSAEncryption)
+
+        XCTAssertEqual(.init(expected.rawRepresentation), found.rawRepresentation)
+    }
+
+    func testEd25519SignatureBytes() throws {
```
Lukasa, line 63: Let's do ECDSA as well?

Lukasa (approved):

PR #265, opened 2025-07-14
Explicit sendable changes

rafaelcepeda:
Performed the necessary changes to enable compiler flag `-require-explicit-sendable`. Also enabled this flag in macOS CI.


Sources/X509/Digests.swift
```diff
@@ -47,6 +47,9 @@ enum Digest {
     }
 }
 
+@available(*, unavailable)
+extension Digest: Sendable {}
```
Lukasa, line 51: This type is intended to be `Sendable`, and only isn't due to the missing Sendability annotations in older API surfaces.

rafaelcepeda: Fixed.

Lukasa (approved):

PR #267, opened 2025-07-17
Improve the performance of `Certificate`'s equal and hash implementations

gjcairo:
Right now, `Certificate` uses the default implementations of `hash` and `==`, which are member-wise.

We can make this more efficient by only comparing the raw bytes for the TBS certificate, the signature, and the signature's algorithm.


Lukasa (approved):
Very nice.

PR #271, opened 2025-07-30
Re-core the benchmarks

Lukasa:
With the release of swift-crypto 3.13.3, our benchmark numbers should have improved. Let's lock those in.


Lukasa:
Look at that, allocation improvements across the board.

glbrntt (approved):

PR #288, opened 2025-12-04
Support generating CMS with trusted signing time

emaloney:
Motivation:

We had a need to generate CMS structures with an embedded trusted timestamp that indicates the time at which a signing operation was performed, and discovered that this was not supported by the existing Swift Certificates implementation.

Modifications:

To address this need, we adapted existing functionality, adding several functions allowing the signing time to be embedded in a CMS structure.

In order to make this work, we also needed to add public access to several entities in the code.

Result:

This new CMS SPI allows Swift Certificates to be used for generating trusted timestamp signatures from CMS structures that embed the signing time, such as those for codesigning.

We have been using this exact same code in a high-volume production environment for several months, and can confirm that the CMS structures generated by this are usable for creating valid signatures.

### DIFF
--- a/Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
+++ b/Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
@@ -80,6 +80,82 @@ public enum CMS: Sendable {
     }
 
     @inlinable
+    static func buildSignedAttributes<Bytes: DataProtocol>(
+        for bytes: Bytes,
+        signatureAlgorithm: Certificate.SignatureAlgorithm,
+        signingTime: Date
+    ) throws -> [CMSAttribute] {
+        // 1. content-type attribute
+        let contentTypeVal = try ASN1Any(erasing: ASN1ObjectIdentifier.cmsData)
+        let contentTypeAttribute = CMSAttribute(attrType: .contentType, attrValues: [contentTypeVal])
+
+        // 2. message-digest attribute
+        let digestAlgorithm = try AlgorithmIdentifier(digestAlgorithmFor: signatureAlgorithm)
+        let computedDigest = try Digest.computeDigest(for: bytes, using: digestAlgorithm)
+        let messageDigest = ASN1OctetString(contentBytes: ArraySlice(computedDigest))
+        let messageDigestVal = try ASN1Any(erasing: messageDigest)
+        let messageDigestAttribute = CMSAttribute(attrType: .messageDigest, attrValues: [messageDigestVal])
+
+        // add signing time utc time in 'YYMMDDHHMMSSZ' format as specificed in `UTCTime`
+        let utcTime = try UTCTime(signingTime.utcDate)
+        let signingTimeAttrVal = try ASN1Any(erasing: utcTime)
+        let signingTimeAttribute = CMSAttribute(attrType: .signingTime, attrValues: [signingTimeAttrVal])
+        return [contentTypeAttribute, messageDigestAttribute, signingTimeAttribute]
+    }
+
+    @_spi(CMS)
+    @inlinable
+    public static func getSignedAttributesBytes<Data: DataProtocol>(
+        digest: Data,
+        signatureAlgorithm: Certificate.SignatureAlgorithm,
+        signingTime: Date
+    ) throws -> [UInt8] {
+        let signedAttrs: [CMSAttribute] = try buildSignedAttributes(
+            for: digest,
+            signatureAlgorithm: signatureAlgorithm,
+            signingTime: signingTime
+        )
+        var coder = DER.Serializer()
+        try coder.serializeSetOf(signedAttrs)
+        return coder.serializedBytes
+    }
+
+    @_spi(CMS)
+    @inlinable
+    public static func signWithTrustedTimestamp<Data: DataProtocol>(
+        _ bytes: Data,
+        signatureBytes: ASN1OctetString,
+        signatureAlgorithm: Certificate.SignatureAlgorithm,
+        additionalIntermediateCertificates: [Certificate] = [],
+        certificate: Certificate,
+        signingTime: Date,
+        detached: Bool = true,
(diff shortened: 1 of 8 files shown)


Sources/X509/X509BaseTypes/AlgorithmIdentifier.swift
```diff
@@ -14,27 +14,23 @@
...
         .sequence
     }
 
-    @usableFromInline
-    var algorithm: ASN1ObjectIdentifier
+    public private(set) var algorithm: ASN1ObjectIdentifier
 
-    @usableFromInline
-    var parameters: ASN1Any?
+    public private(set) var parameters: ASN1Any?
```
Lukasa: I don't think there's any particular reason to make these `private(set)`.

Sources/X509/CryptographicMessageSyntax/CMSEncapsulatedContentInfo.swift
```diff
@@ -21,10 +21,9 @@ import SwiftASN1
 ///   eContent [0] EXPLICIT OCTET STRING OPTIONAL }
 /// ContentType ::= OBJECT IDENTIFIER
 /// ```
-@usableFromInline
-struct CMSEncapsulatedContentInfo: DERImplicitlyTaggable, BERImplicitlyTaggable, Hashable, Sendable {
+public struct CMSEncapsulatedContentInfo: DERImplicitlyTaggable, BERImplicitlyTaggable, Hashable, Sendable {
```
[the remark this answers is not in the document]
emaloney: This change is not present in the latest version of this PR.

Sources/X509/CryptographicMessageSyntax/CMSVersion.swift
```diff
@@ -19,27 +19,25 @@ import SwiftASN1
 ///  CMSVersion ::= INTEGER
 ///                 { v0(0), v1(1), v2(2), v3(3), v4(4), v5(5) }
 /// ```
-@usableFromInline
-struct CMSVersion: RawRepresentable, Hashable, Sendable {
-    @usableFromInline
-    var rawValue: Int
+public struct CMSVersion: RawRepresentable, Hashable, Sendable {
```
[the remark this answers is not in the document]
emaloney: This change is not present in the latest version of this PR.

Sources/X509/X509BaseTypes/AlgorithmIdentifier.swift
emaloney: Changed to just plain `public` in latest version of this PR.

Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
```diff
@@ -78,39 +78,106 @@ public enum CMS: Sendable {
...
         let utcTime = try UTCTime(signingTime.utcDate)
         let signingTimeAttrVal = try ASN1Any(erasing: utcTime)
         let signingTimeAttribute = CMSAttribute(attrType: .signingTime, attrValues: [signingTimeAttrVal])
-        signedAttrs.append(signingTimeAttribute)
+        return [contentTypeAttribute, messageDigestAttribute, signingTimeAttribute]
+    }
+
+    @_spi(CMS)
+    @inlinable
+    public static func createSigningTimeASN1(signingTime: Date) throws -> Data {
```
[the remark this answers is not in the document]
emaloney: This change is not present in the latest version of this PR.

Sources/X509/Digests.swift
```diff
@@ -19,16 +19,15 @@ import Foundation
 #endif
 @preconcurrency import Crypto
 
-@usableFromInline
 @available(macOS 10.15, iOS 13, watchOS 6, tvOS 13, macCatalyst 13, visionOS 1.0, *)
-enum Digest: Sendable {
+public enum Digest: Sendable {
```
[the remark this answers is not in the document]
emaloney: I see that since my original PR was opened, `Digest` has gone from an `enum` to a `struct`, and the cases you were (presumably) worried about exposing are now hidden in the inner `enum` type `Backing`.

Am I correct in assuming that making `Digest` and the `computeDigest(for:using:)` function `public` is no longer a concern given this refactoring?

emaloney:
@Lukasa I've addressed the earlier PR comments & rebased on top of the latest `swift-certificates`. In the process of adding unit tests, I noticed an apparent bug (issue #306) that would need to be addressed first (PR #307).

In the meantime, I rebased #288 on top of #307 so the new tests would pass. Once #307 gets merged, I'll rebase this on the latest `main` again, and this PR should be ready for (final?) re-review.

Sources/X509/CryptographicMessageSyntax/CMSOperations.swift
```diff
@@ -80,39 +80,96 @@ public enum CMS: Sendable {
...
         let utcTime = try UTCTime(signingTime.utcDate)
         let signingTimeAttrVal = try ASN1Any(erasing: utcTime)
         let signingTimeAttribute = CMSAttribute(attrType: .signingTime, attrValues: [signingTimeAttrVal])
-        signedAttrs.append(signingTimeAttribute)
+        return [contentTypeAttribute, messageDigestAttribute, signingTimeAttribute]
+    }
 
+    @_spi(CMS)
+    @inlinable
+    public static func getSignedAttributesBytes<Data: DataProtocol>(
```
Lukasa, line 107: Is there a reason this needs to be an SPI?

emaloney: No technical reason; I will remove that.

emaloney: Pushed an update that removes the `@_spi(CMS)` — note that the `signWithTrustedTimestamp` function is still marked as SPI. I left that as-is b/c I wasn't sure if that signature was something we were ready to formalize & make available publicly.

PR #296, opened 2026-04-27
documentation updates

heckj:
- **Fix stale DocC disambiguation hash for PrivateKey init**
- **Fix documentation errors, typos, and a bug in CertificateStore.appending**
- **code example fixes within the existing articles:**
- **Lots of smaller fixes across the board - some copy/pasta errors, consistency in usage, typos and grammar issues - added documentation comments where missing to get all public types covered - applied internal style guidelines for representing the types**


Sources/X509/Verifier/CertificateStore.swift
```diff
@@ -66,7 +66,7 @@ public struct CertificateStore: Sendable, Hashable {
     public func appending(_ certificate: Certificate) -> Self {
```
dnadoba: ouch, good catch. 
We can upgrade this now as well to just use consuming, get rid of the copy and make this kind of bug impossible:
```suggestion
    consuming func appending(_ certificate: Certificate) -> Self {
        append(certificate)
        return self
    }
```
(we couldn't use that before as it wasn't available in the swift version we needed to support)

dnadoba (commented):
A simple test for that would be great as well but I wouldn't say required.
Looked briefly over the rest and it looks okay but I will leave that up to the new maintainers.

heckj:
In the method that I found an issue in, I can just fix the bug, or I can use "consuming" either public or internal as @dnadoba suggested, but that changes the API surface and would seem to warrant a semver/patch or major change - depending on what this project would prefer. 

I'm happy to go in which-ever direction is desired. I meant to only cover documentation updates and stumbled across the issue when checking for consistency issues.

dnadoba:
For copyable types, like in this case, this isn't API breaking. I wouldn't even consider it an API change as this should really just be a calling convention/optimization difference.

Sources/X509/Docs.docc/Examining Certificates.md
```diff
@@ -2,15 +2,15 @@
 
 Decoding and introspecting certificates.
 
-A common use-case is to have received an X.509 certificate from some source, and to want
-to inspect it and make decisions based on its content. X.509 certificates are complex
-objects with highly dynamic and nested content, so this article attempts to outline
-how to work with ``Certificate`` in order to get the most out of the content of an
+A common use case is to receive an X.509 certificate from some source and
+inspect it to make decisions based on its content. X.509 certificates are complex
+objects with highly dynamic and nested content, so this article outlines
```
josephnoir:
```suggestion
objects with highly dynamic and nested content. This article outlines
```

Sources/X509/Docs.docc/Examining Certificates.md
```diff
@@ -30,20 +30,20 @@ FGgfAiEA1PMpcTTeK8Hi17N0oaNNIWERLqDketPZG5E3OdEDw9E=
 ```
 
 This format is the DER representation of a certificate, base64-encoded, line wrapped at 64
-bytes, and bracketed by the `BEGIN X`/`END X` delimiter. As PEMs can easily be translated
+bytes, and bracketed by the `BEGIN X`/`END X` delimiters. As PEMs can easily be translated
```
josephnoir, line 33: Stumbling over this one. (Might be right nonetheless.)

Lukasa (approved):
Thanks @heckj!

heckj:
The API breakage check is the switching up to using `consuming` on the API - that's expected. The last commit I added in resolved the broken disambiguation hashes that work on macOS, but didn't when built on Linux.

josephnoir (approved):
I like it. Thanks! I got a nit and I stumbled over the singular of "delimiters". I'll let you be the judge though.

Sources/X509/Docs.docc/Examining Certificates.md
heckj: It read more plural to me, but if you really don't like it I can switch it up.

PR #302, opened 2026-06-30
Add env flag to accept swift-crypto beta releases

josephnoir:
### Motivation

Tagged releases can only have other tagged releases in their dependencies. Tagging a package that depends on a swift-crypto beta and swift-certificates requires a swift-certificates release that accepts a swift-crypto beta release.

### Modifications

Introduce an environment flag that allows swift-certificates to depend on swift-crypto beta releases.

### Results

Setting an environment flag can widen the upper limit of accepted swift-cryto version to include 5.0.0 beta release.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -74,9 +74,25 @@ let package = Package(
 // we're building in the Swift.org CI system alongside other projects in the Swift toolchain and
 // we can depend on local versions of our dependencies instead of fetching them remotely.
 if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
+
+    // If the `SWIFT_CERTIFICATES_ALLOW_SWIFT_CRYPTO_BETA` environment variable is set
+    // swift-certificates will accept swift-crypto beta releases as a dependency.
+    //
+    // Note: A beta release can only be used if other packages in the dependency tree
+    // that have a direct dependency on swift-crypto accept beta releases as well.
+    if ProcessInfo.processInfo.environment["SWIFT_CERTIFICATES_ALLOW_SWIFT_CRYPTO_BETA"] == nil {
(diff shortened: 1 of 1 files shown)


Package.swift
```diff
@@ -74,9 +74,25 @@ let package = Package(
 // we're building in the Swift.org CI system alongside other projects in the Swift toolchain and
 // we can depend on local versions of our dependencies instead of fetching them remotely.
 if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] == nil {
+
+    // If the `SWIFT_CERTIFICATES_ALLOW_SWIFT_CRYPTO_BETA` environment variable is set
+    // swift-certificates will accept swift-crypto beta releases as a dependency.
+    //
+    // Note: A beta release can only be used if other packages in the dependency tree
+    // that have a direct dependency on swift-crypto accept beta releases as well.
+    if ProcessInfo.processInfo.environment["SWIFT_CERTIFICATES_ALLOW_SWIFT_CRYPTO_BETA"] == nil {
```
Lukasa, line 83: Is this intentionally nested?

josephnoir: Yes, the intention was for local dependencies to have precedence. Flipping the comparisons could make this an `else if` branch.

```swift
if ProcessInfo.processInfo.environment["SWIFTCI_USE_LOCAL_DEPS"] != nil {
  // use local repos
} else if ProcessInfo.processInfo.environment["SWIFT_CERTIFICATES_ALLOW_SWIFT_CRYPTO_BETA"] != nil {
  // accept swift-crypto beta releases
} else {
  // default dependency range
}
```

Lukasa: Oh gotcha, I see.

Lukasa (approved):

PR #312, opened 2026-08-11
Add ML-DSA-65 and ML-DSA-87 certificate parsing and verification

justsem:
swift-certificates can't yet parse a certificate signed with ML-DSA: it throws on the issuer's signature algorithm before verification even starts, even though swift-crypto already supports these algorithms. That turned out to be a blocker in one of our projects involving PQ PKI.

This PR adds parsing and verification for ML-DSA-65 and ML-DSA-87, per RFC 9881.

- First commit bumps swift-crypto to 4.0.0, where MLDSA65/MLDSA87 was introduced.
- Certificates signed by an ML-DSA CA now parse and round-trip byte-for-byte. ML-DSA public keys work through the usual Certificate.PublicKey API, and chains validate through Verifier with RFC5280Policy.
- ML-DSA requires macOS/iOS 26 at runtime (since it comes from CryptoKit there). Below that, decoding an ML-DSA key fails with an error.

### DIFF
--- a/Benchmarks/Package.swift
+++ b/Benchmarks/Package.swift
@@ -23,7 +23,7 @@ let package = Package(
     dependencies: [
         .package(path: "../"),
         .package(url: "https://github.com/ordo-one/benchmark.git", from: "1.11.1"),
-        .package(url: "https://github.com/apple/swift-crypto.git", "4.0.0"..<"5.0.0"),
+        .package(url: "https://github.com/apple/swift-crypto.git", "3.12.3"..<"5.0.0"),
         .package(url: "https://github.com/apple/swift-asn1.git", from: "1.0.0"),
     ],
     targets: [
--- a/Benchmarks/Package.swift
+++ b/Benchmarks/Package.swift
@@ -23,7 +23,7 @@ let package = Package(
     dependencies: [
         .package(path: "../"),
         .package(url: "https://github.com/ordo-one/benchmark.git", from: "1.11.1"),
-        .package(url: "https://github.com/apple/swift-crypto.git", "3.12.3"..<"5.0.0"),
+        .package(url: "https://github.com/apple/swift-crypto.git", "4.0.0"..<"5.0.0"),
         .package(url: "https://github.com/apple/swift-asn1.git", from: "1.0.0"),
     ],
     targets: [
--- a/Sources/X509/CertificatePublicKey.swift
+++ b/Sources/X509/CertificatePublicKey.swift
@@ -309,8 +309,8 @@ extension SubjectPublicKeyInfo {
             algorithmIdentifier = .ed25519
             key = .init(bytes: ArraySlice(ed25519.rawRepresentation))
         case .mldsa(let mldsa):
-            algorithmIdentifier = (mldsa.variant == .mldsa65) ? .mldsa65 : .mldsa87
-            key = .init(bytes: ArraySlice(mldsa.bytes))
+            algorithmIdentifier = AlgorithmIdentifier(mldsaVariant: mldsa.variant)
+            key = .init(bytes: ArraySlice(mldsa.rawRepresentation))
         }
 
         self.algorithmIdentifier = algorithmIdentifier
--- a/Sources/X509/CertificatePublicKey.swift
+++ b/Sources/X509/CertificatePublicKey.swift
@@ -54,6 +54,10 @@ extension Certificate {
             case .ed25519:
                 let key = try Curve25519.Signing.PublicKey(rawRepresentation: spki.key.bytes)
                 self.backing = .ed25519(key)
+            case .mldsa65:
+                self.backing = .mldsa(try MLDSAPublicKeyBytes(spkiBytes: spki.key.bytes, variant: .mldsa65))
+            case .mldsa87:
+                self.backing = .mldsa(try MLDSAPublicKeyBytes(spkiBytes: spki.key.bytes, variant: .mldsa87))
             default:
                 throw CertificateError.unsupportedPublicKeyAlgorithm(reason: "\(spki.algorithmIdentifier)")
             }
@@ -154,6 +158,8 @@ extension Certificate.PublicKey {
             return rsa.isValidSignature(signature, for: bytes, signatureAlgorithm: signatureAlgorithm)
         case .ed25519(let ed25519):
             return ed25519.isValidSignature(signature, for: bytes, signatureAlgorithm: signatureAlgorithm)
+        case .mldsa(let mldsa):
+            return mldsa.isValidSignature(signature, for: bytes, signatureAlgorithm: signatureAlgorithm)
         }
     }
 
@@ -185,6 +191,8 @@ extension Certificate.PublicKey {
             return rsa.isValidSignature(signature, for: bytes, signatureAlgorithm: signatureAlgorithm)
         case .ed25519(let ed25519):
             return ed25519.isValidSignature(signature, for: bytes, signatureAlgorithm: signatureAlgorithm)
+        case .mldsa(let mldsa):
+            return mldsa.isValidSignature(signature, for: bytes, signatureAlgorithm: signatureAlgorithm)
         }
     }
 }
@@ -209,6 +217,13 @@ extension Certificate.PublicKey: CustomStringConvertible {
             return "RSA\(publicKey.keySizeInBits).PublicKey"
         case .ed25519:
             return "Ed25519.PublicKey"
+        case .mldsa(let backing):
+            switch backing.variant {
+            case .mldsa65:
+                return "MLDSA65.PublicKey"
+            case .mldsa87:
+                return "MLDSA87.PublicKey"
+            }
         }
     }
 }
@@ -222,6 +237,7 @@ extension Certificate.PublicKey {
         case p521(Crypto.P521.Signing.PublicKey)
         case rsa(_CryptoExtras._RSA.Signing.PublicKey)
         case ed25519(Curve25519.Signing.PublicKey)
+        case mldsa(MLDSAPublicKeyBytes)
 
         @inlinable
         static func == (lhs: BackingPublicKey, rhs: BackingPublicKey) -> Bool {
@@ -236,6 +252,8 @@ extension Certificate.PublicKey {
                 return l.derRepresentation == r.derRepresentation
             case (.ed25519(let l), .ed25519(let r)):
                 return l.rawRepresentation == r.rawRepresentation
+            case (.mldsa(let l), .mldsa(let r)):
+                return l == r
             default:
                 return false
             }
@@ -259,6 +277,9 @@ extension Certificate.PublicKey {
             case .ed25519(let digest):
                 hasher.combine(4)
                 hasher.combine(digest.rawRepresentation)
+            case .mldsa(let backing):
+                hasher.combine(5)
+                hasher.combine(backing)
             }
         }
     }
@@ -287,6 +308,9 @@ extension SubjectPublicKeyInfo {
         case .ed25519(let ed25519):
             algorithmIdentifier = .ed25519
             key = .init(bytes: ArraySlice(ed25519.rawRepresentation))
+        case .mldsa(let mldsa):
+            algorithmIdentifier = (mldsa.variant == .mldsa65) ? .mldsa65 : .mldsa87
+            key = .init(bytes: ArraySlice(mldsa.bytes))
         }
 
         self.algorithmIdentifier = algorithmIdentifier
--- a/Sources/X509/MLDSA.swift
+++ b/Sources/X509/MLDSA.swift
@@ -101,7 +101,11 @@ extension MLDSAPublicKeyBytes {
         }
         guard #available(macOS 26.0, iOS 26.0, watchOS 26.0, tvOS 26.0, macCatalyst 26.0, visionOS 26.0, *)
         else {
-            return false
+            // Every initializer of MLDSAPublicKeyBytes requires ML-DSA availability, so no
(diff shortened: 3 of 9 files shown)


Benchmarks/Package.swift
```diff
@@ -23,7 +23,7 @@ let package = Package(
     dependencies: [
         .package(path: "../"),
         .package(url: "https://github.com/ordo-one/benchmark.git", from: "1.11.1"),
-        .package(url: "https://github.com/apple/swift-crypto.git", "3.12.3"..<"5.0.0"),
+        .package(url: "https://github.com/apple/swift-crypto.git", "4.0.0"..<"5.0.0"),
```
Joannis: I was discussing using a package trait to avoid needing this, let us know what works.

Joannis:
CC @fabianfett if you're back, could we get a CI run on this?

Sources/X509/MLDSA.swift
```diff
@@ -0,0 +1,187 @@
...
+    ) -> Bool {
+        switch (self.variant, signatureAlgorithm) {
+        case (.mldsa65, .mldsa65), (.mldsa87, .mldsa87):
+            break
+        default:
+            return false
+        }
+        guard #available(macOS 26.0, iOS 26.0, watchOS 26.0, tvOS 26.0, macCatalyst 26.0, visionOS 26.0, *)
+        else {
+            return false
```
Lukasa: Is this code branch actually reachable? It seems as though it should not be possible to create a key on these platforms such that we'd be unable to get here.

justsem: You're correct, it's not reachable. But the compiler required it in the way the PR is made now. I'd be happy to change it up, but this seemed like the best way.

Lukasa: Oh I meant more that it should probably be a `fatalError` with the justification encoded.

justsem: I've changed it to a fatal error :)

Sources/X509/MLDSA.swift
```diff
@@ -0,0 +1,191 @@
...
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of SwiftCertificates project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import SwiftASN1
```
aryan-25: Doesn't look like this import is being used in this file.

Sources/X509/MLDSA.swift
```diff
@@ -0,0 +1,191 @@
...
+/// and rebuilt on each use. The payload is plain data so the enum case that carries it can
+/// be declared without an availability annotation (macOS-26-only types can't appear in an
+/// unannotated enum payload); only the *initializers* need swift-crypto's ML-DSA API.
+@usableFromInline
+struct MLDSAPublicKeyBytes: Hashable, Sendable {
+    @usableFromInline
+    var variant: MLDSAVariant
+
+    @usableFromInline
+    var bytes: Data
```
aryan-25: This property should perhaps be renamed to `rawRepresentation` to be consistent with `swift-crypto`'s ML-DSA public key type.

Sources/X509/CertificatePublicKey.swift
```diff
@@ -287,6 +308,9 @@ extension SubjectPublicKeyInfo {
         case .ed25519(let ed25519):
             algorithmIdentifier = .ed25519
             key = .init(bytes: ArraySlice(ed25519.rawRepresentation))
+        case .mldsa(let mldsa):
+            algorithmIdentifier = (mldsa.variant == .mldsa65) ? .mldsa65 : .mldsa87
```
aryan-25: nit: I think this should be extracted out into an extension on `AlgorithmIdentifier`:

```
extension AlgorithmIdentifier {
    @usableFromInline
    init(mldsaVariant: MLDSAVariant) {
        switch mldsa {
        case .mldsa65:
            self = .mldsa65

        case .mldsa87:
            self = .mldsa87
        }
    }
}
```

Sources/X509/MLDSA.swift
```diff
@@ -0,0 +1,191 @@
...
+#else
+import Foundation
+#endif
+
+/// The ML-DSA parameter sets this library can name.
+///
+/// This type deliberately stores no key material and names no swift-crypto ML-DSA type,
+/// so it can be declared without an availability annotation: enum cases can't carry
+/// `@available`, so keeping this payload free of macOS-26-only types is what lets
+/// certificates *signed* with ML-DSA be parsed and re-serialized unconditionally.
```
aryan-25: I think these lines are unnecessary in the context of this type and can be removed.

Sources/X509/MLDSA.swift
```diff
@@ -0,0 +1,191 @@
...
+            break
+        default:
+            return false
+        }
+        guard #available(macOS 26.0, iOS 26.0, watchOS 26.0, tvOS 26.0, macCatalyst 26.0, visionOS 26.0, *)
+        else {
+            // Every initializer of MLDSAPublicKeyBytes requires ML-DSA availability, so no
+            // value of this type can exist on a platform that fails this check. The guard is
+            // only here because this function must be callable from baseline-availability
+            // dispatch code while the swift-crypto calls below are macOS 26+ on Darwin.
```
aryan-25:
```suggestion
            // The initializer validates the platform requirement, so this `fatalError` is unreachable.
```

Sources/X509/MLDSA.swift
```diff
@@ -0,0 +1,191 @@
...
+    case mldsa87
+}
+
+/// The raw bytes of an ML-DSA public key, plus its parameter set.
+///
+/// Reconstructing a swift-crypto ML-DSA public key from its raw representation costs ~0.4%
+/// of a verification (measured in the design doc), so the key is stored as validated bytes
+/// and rebuilt on each use. The payload is plain data so the enum case that carries it can
+/// be declared without an availability annotation (macOS-26-only types can't appear in an
+/// unannotated enum payload); only the *initializers* need swift-crypto's ML-DSA API.
```
aryan-25: This can be expressed more concisely. The only two important things to note are:

1. This type cannot store swift-crypto's `MLDSA65.PublicKey` / `MLDSA87.PublicKey` while being available on all platforms, because `swift-crypto`'s ML-DSA APIs have 26.0 platform requirements.
2. As such, only the key's bytes are stored. Methods that require swift-crypto's MLDSA APIs create an `MLDSA65.PublicKey` / `MLDSA87.PublicKey` instance from the raw bytes.

Tests/X509Tests/MLDSATests.swift
```diff
@@ -0,0 +1,893 @@
...
+import SwiftASN1
+@preconcurrency import Crypto
+@testable import X509
+
+final class MLDSATests: XCTestCase {
+    // MARK: - No ML-DSA key material involved: parsing and re-serializing the ASN.1 identity layer
+
+    // A P-256 leaf signed by an ML-DSA-65 CA. Committed in wendylabs cloud-ws-c as a canary
+    // asserting swift-certificates could NOT parse it; generated with OpenSSL 3.6.3
+    // (`openssl genpkey -algorithm ML-DSA-65`). Interop evidence from a second implementation.
```
aryan-25: These lines can be removed. The name of the test function is explanatory.

Tests/X509Tests/MLDSATests.swift
```diff
@@ -0,0 +1,893 @@
...
+            XCTAssertEqual((error as? CertificateError)?.code, .unsupportedPublicKeyAlgorithm)
+        }
+    }
+
+    // MARK: - Requires macOS 26+ on Darwin: names MLDSA65/MLDSA87 and touches key material
+
+    func testParsesSelfSignedMLDSA65Root() throws {
+        guard #available(macOS 26.0, iOS 26.0, watchOS 26.0, tvOS 26.0, macCatalyst 26.0, visionOS 26.0, *) else {
+            throw XCTSkip("ML-DSA requires macOS 26 / iOS 26 on Darwin")
+        }
```
aryan-25: Note that you can annotate the test case with `@available` and avoid the `#available` check within the test: 
```suggestion
    @available(macOS 26.0, iOS 26.0, watchOS 26.0, tvOS 26.0, macCatalyst 26.0, visionOS 26.0, *)
    func testParsesSelfSignedMLDSA65Root() throws {
```

The same applies for all test cases below.

Tests/X509Tests/MLDSATests.swift
```diff
@@ -0,0 +1,893 @@
...
+        XCTAssertEqual(Array(der), serializer.serializedBytes)
+    }
+
+    // ML-DSA-44 exists in RFC 9881 (2.16.840.1.101.3.4.3.17) but not in swift-crypto,
+    // so we deliberately do not support it. The fixture is a self-signed ML-DSA-44 root, so
+    // its OWN subjectPublicKeyInfo (also id-ml-dsa-44) is decoded first, while parsing the
+    // TBSCertificate — before the outer signatureAlgorithm/signatureValue are ever reached.
+    // That decode order means this always surfaces as unsupportedPublicKeyAlgorithm, not
+    // unsupportedSignatureAlgorithm; verified empirically against Sources/X509/Certificate.swift
+    // and CertificatePublicKey.swift's `init(spki:)`.
```
aryan-25:
```suggestion
    // ML-DSA-44 exists in RFC 9881 (2.16.840.1.101.3.4.3.17) but not in swift-crypto,
    // so we deliberately do not support it.
```

Tests/X509Tests/MLDSATests.swift
```diff
@@ -0,0 +1,893 @@
...
+    func testRejectsChainToWrongRoot() async throws {
+        guard #available(macOS 26.0, iOS 26.0, watchOS 26.0, tvOS 26.0, macCatalyst 26.0, visionOS 26.0, *) else {
+            throw XCTSkip("ML-DSA requires macOS 26 / iOS 26 on Darwin")
+        }
+        // An "evil twin" root: same subject DN as the real ML-DSA-65 root, but a different
+        // key. swift-certificates looks up candidate issuers by subject DN alone (AKI/SKI
+        // only affects sort *preference* among several candidates, not eligibility -- see
+        // Verifier.issuerPreference), so this twin is genuinely considered as an issuer for
+        // the leaf below, and is rejected only because the leaf's SIGNATURE does not verify
+        // under its key -- not because of a name/DN mismatch.
```
aryan-25:
```suggestion
        // An "evil twin" root: same subject DN as the real ML-DSA-65 root, but a different
        // key.
```

Tests/X509Tests/MLDSATests.swift
```diff
@@ -0,0 +1,893 @@
...
+// CertificateStore.ConcreteBacking.additionalTrustRoots, keyed by Certificate.subject), so
+// this twin IS selected as a candidate issuer when chain-building looks for an issuer of
+// the real root's leaf -- it is rejected only because the leaf's SIGNATURE does not verify
+// under its (different) key, not because of a name/DN mismatch. Its SKI necessarily differs
+// from the real root's too (SKI is `hash` of its own public key), which only affects
+// Verifier.issuerPreference's *sort order* among candidates, not eligibility, so the
+// signature-verification path is still exercised. Generated with OpenSSL 3.6.3:
+//   openssl req -x509 -new -newkey ML-DSA-65 -nodes -keyout evil-twin.key \
+//     -out evil-twin.pem -days 3650 -subj "/CN=swift-certificates ML-DSA-65 Test Root"
+//   openssl x509 -in evil-twin.pem -outform DER -out evil-twin.der
```
aryan-25, line 767:
```suggestion
// An "evil twin" of the ML-DSA-65 root above: same subject DN, a brand-new ML-DSA-65 key,
// self-signed. Generated with OpenSSL 3.6.3:
//   openssl req -x509 -new -newkey ML-DSA-65 -nodes -keyout evil-twin.key \
//     -out evil-twin.pem -days 3650 -subj "/CN=swift-certificates ML-DSA-65 Test Root"
//   openssl x509 -in evil-twin.pem -outform DER -out evil-twin.der
```

Tests/X509Tests/MLDSATests.swift
```diff
@@ -0,0 +1,893 @@
...
+private func fixtureCertificate(_ base64: String) throws -> Certificate {
+    let der = try XCTUnwrap(Data(base64Encoded: base64))
+    return try Certificate(derEncoded: Array(der))
+}
+
+// A P-256 leaf signed by an ML-DSA-65 CA. Reused verbatim (base64 content only) from
+// wendylabs cloud-ws-c's `RequestSignatureVerifierTests.swift`, where it is committed as an
+// interop canary asserting swift-certificates could NOT parse it. Generated there with
+// OpenSSL 3.6.3 (`openssl genpkey -algorithm ML-DSA-65` for the CA, then a P-256 CSR signed
+// by that CA). Interop evidence from a second implementation.
```
aryan-25:
```suggestion
// A P-256 leaf signed by an ML-DSA-65 CA. Generated with OpenSSL 3.6.3
// (`openssl genpkey -algorithm ML-DSA-65` for the CA, then a P-256 CSR signed
// by that CA).
```

aryan-25 (changes requested):

Benchmarks/Package.swift
Lukasa: I think a package trait will be necessary for this, unfortunately.

Lukasa:
There are also some test failures that need looking at.

Joannis:
@Lukasa can you re-trigger CI?

PR #313, opened 2026-08-15
Make _TinyArray compatible with library evolution

Anywhere-Music-Player:
### Motivation:

Building `_CertificateInternals` with library evolution enabled fails because `_TinyArray` and its `@usableFromInline` storage are referenced from `@inlinable` code while their layout and cases are not fixed.

### Modifications:

- Marked `_TinyArray` as `@frozen`.
- Marked its nested `Storage` enum as `@frozen`.

### Result:

`_CertificateInternals` now builds with library evolution enabled in both Swift 5 and Swift 6 language modes.


Anywhere-Music-Player:
Hi @Lukasa and @josephnoir, could you review this two-line fix for library evolution builds? The CI runs expired while waiting for approval. Could you approve a rerun and add the `🔨 semver/patch` label? Thanks!

Lukasa (approved):
Thanks, this is a reasonable change.

PR #317, opened 2026-09-06
Treat a /0 iPAddress name constraint as covering the whole address space

lenamonj:
### Motivation

`NameConstraintsPolicy` read an all-zero CIDR mask as matching nothing. An all-zero mask constrains no bits, so it matches every address: the CA/Browser Forum Baseline Requirements express "this subordinate CA must not issue for IP addresses" as `excludedSubtrees` of `0.0.0.0/0` and `::/0`, and against this code those exclusions constrained nothing, so a leaf with an IP SAN under such a CA verified.

### Modifications

The leading-zero shortcut in `isValidCIDRMask` is removed and the general leading-ones-then-zeros check decides. The two `IPAddressNameTests` rows that pinned the old reading now expect a match. `testAllZeroMaskIsValidCIDRMask` asserts the IPv4 and IPv6 all-zero masks directly, and two `RFC5280PolicyTests1` tests run `RFC5280Policy` with `0.0.0.0/0` and `::/0` as permitted and as excluded subtrees. All of them fail on main.

### Result

A `/0` iPAddress constraint covers the whole address space, as RFC 5280 4.2.1.10 defines it. `swift test` passes under the CI flags (`-Xswiftc -warnings-as-errors --explicit-target-dependency-import-check error -Xswiftc -require-explicit-sendable`) on Swift 6.1.2; `swift-format lint --strict` is clean on the changed files.


josephnoir:
Thank you for the contribution! This looks like the right fix. It would be great if you could add two more test cases:

1. An explicit test that all-zero masks pass `isValidCIDRMask` (ipv4/ipv6).
2. Tests for the `RFC5280Policy` that use an all-zero mask for name constraints with ip addresses (permitted/excluded).

lenamonj:
Both are in 9fb536f. `testAllZeroMaskIsValidCIDRMask` asserts `isValidCIDRMask` for the IPv4 and IPv6 all-zero masks, and two `RFC5280PolicyTests1` tests run `RFC5280Policy` with `0.0.0.0/0` and `::/0` as permitted and as excluded subtrees. All three fail on main.

josephnoir (approved):
LGTM, thanks!

PR #319, opened 2026-09-17
Accept Swift Crypto 5 final release

fpseverino:
Motivation:

The final release of Swift Crypto 5 is now available, so we don't need the `SWIFT_CERTIFICATES_ALLOW_SWIFT_CRYPTO_BETA` env flag anymore and we can accept the version range `"3.12.3"..<"6.0.0"`

Modifications:

Remove the `SWIFT_CERTIFICATES_ALLOW_SWIFT_CRYPTO_BETA` env var logic and accept Swift Crypto `"3.12.3"..<"6.0.0"`

Result:

Projects can require both the final release of Swift Crypto 5 and the latest Swift Certificates release


Lukasa (approved):
Nice cleanup, thanks @fpseverino!

